PR #7, opened 2019-10-08
KIP 0002: fungible-v1

slpopejoy:
See <https://github.com/kadena-io/KIPs/blob/kip-0002-proposal/kip-0002.md> and code in <https://github.com/kadena-io/KIPs/tree/kip-0002-proposal/kip-0002>.


emilypi:
Regarding `TODO: should create-account check the guard?`

There are two arguments I see for and against this: 

1. The argument to the negative would be that this is a convenience check which everyone pays for. This should be enforced by wallets, and otherwise leave users to check their credentials before requesting an account, lest they pay to lock themselves out of a dead account.

2. The contrapositive would be that there is little incentive for us as developers to allow people to potentially grief our user tables by spamming DOA accounts and increasing DB latency for everyone on the network. 

I'm more in favor of the "correct by construction" design, which makes sure that only well-formed accounts (including their guards) are allowed to participate. 

With respect to `TODO: should rotate check the new guard too?`:

The same arguments apply.

slpopejoy:
To be clear, anything left out because `interface` does not support it is simply to get this over the line. If we add support in interfaces in time then happy to include. These include:

- schema for `info` results

- defcap for TRANSFER managed cap

etc

slpopejoy:
It looks like we will have the necessary support. Therefore, proposing we add

```
(defschema account-info 
  "Details of account."
  account:string
  balance:decimal 
  guard:guard)

(defpact move 
    ( sender:string
      receiver:string
      receiver-guard:guard-chain-id:string
      target-chain:string
      amount:decimal ))

(defcap TRANSFER 
    ( sender:string
      receiver:string
      amount:decimal
      )
      @managed TRANSFER-mgr)

(defun TRANSFER-mgr (mgd rec))
```

slpopejoy:
Also, we can consider an on-behalf-of functionality. Transfer on behalf of allows a sender to delegate payments to a third party, like an exchange. See https://programtheblockchain.com/posts/2018/01/30/writing-an-erc20-token-contract/#delegated-transfer-functionality .  

It's not clear why this can't simply be a user guard where the sender creates a multisig account for the exchange. 

1. Sender transfers balance to new multisig account [sender,delegate] _ud_.
2. Delegate makes transfers from _ud_ to whoever.
3. Sender at any time can drain account back to original account; sender can conversely fund the account. 
4. Delegate can refund back to shared account at any time.

What's missing here? Doesn't seem like much, unless there's some need for the user to guarantee funds availability for some time, which becomes a user guard with a timeout before the sender can mess with it:

```
(defun on-behalf-guard ( sender-guard:guard delegate-guard:guard expiry:time )
  (enforce-one [
    (enforce-guard delegate-guard)
    (and (enforce-guard sender-guard)
      (enforce (> expiry (at 'block-time (chain-data)))
        (format "Expiry not reached: {}" [expiry])))]))

(create-user-guard (on-behalf-guard sender-guard delegate-guard expiry))
```

Therefore it doesn't seem clear that `fungible` needs special instrumentation for this.

slpopejoy:
added TRANSFER, transfer-chain, transfer-schema, account-details, updated enforce-unit type, balance->get-balance.

added full reference impl.

emilypi:
I've updated the KIP to reflect the most recent discussion, enforcing guards on account creation and guard rotation, and added invariants to certain schema

kip-0002/fungible-v1.pact
```diff
@@ -0,0 +1,133 @@
...
+     @doc " Manages TRANSFER capability. REQUESTED should match MANAGED on \
+          \ sender and receiver. AMOUNT should be validated against managed, \
+          \ and linearly managed such that a request for 1.0 amount on a 3.0 \
+          \ managed quantity emits MANAGED with updated amount 2.0."
+     )
+
+   ; ----------------------------------------------------------------------
+   ; Functionality
+
+   (defun transfer-create:string
```
emilypi, line 55: This needs a standalone `transfer` function as well

slpopejoy: That will be in the KIP for `-v2`

PR #11, opened 2020-08-14
KIP-0006: Standard Token Transfer Process

mightybyte:
---
  - https://app.asana.com/0/0/1208668844268840


edmundnoble:
The first KIP council failed to comment on this KIP the first time, and we resolve to return to this KIP later after doing more research.

javadkh2:
This is how we handle account discovery and transferring assets in chainweaver 3 
Untitled-2025-07-16-1536

edmundnoble:
This KIP is somewhat outdated now due to the introduction of principal accounts, which add a new way for a guard to match an account name, beyond just having the account name equal its controlling public key. Chainweaver 3 has adopted a similar transfer process, and its design is posted above.

The second KIP council resolves to close this KIP for the time being, but if it were updated to bring it in line with the latest features of Pact and design of Chainweaver 3, it might be merged in future.

PR #17, opened 2021-05-21
KIP-0010: Fungible regression test

sirlensalot:
Not ready for comment yet

---
  - https://app.asana.com/0/0/1208668844268848

### DIFF
--- a/kip-0010.md
+++ b/kip-0010.md
@@ -1,0 +1,50 @@
+---
+KIP: 0010
+Title: fungible-v2 Event Standard
+Author: Stuart Popejoy stuart@kadena.io
+Status: Draft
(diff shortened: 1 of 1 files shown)


edmundnoble:
The KIP council resolved in KIP Council 1 that this KIP:
- has value in documenting and testing fungible token contracts, which we've discussed doing more in the past.
- is just a draft, and very incomplete.
- would not require a lot of work, mostly just altering the existing tests for the `coin` contract.

Thus we have added it to an internal backlog. We have also added tasks for further documenting the process of authoring new fungible token contracts. We will be moving this to the Issues section as a motivator for future work and KIPs.

PR #22, opened 2021-12-22
KIP-0014: Updated KIP-0012 Account Protocols

sirlensalot:
Augment KIP-0012 account protocols with two new protocols, Guard Accounts and Rotatable Accounts.
Modify KIP-0012 Single-Key Accounts to forbid rotation.

### DIFF
--- a/kip-0014/kip-0014.md
+++ b/kip-0014/kip-0014.md
@@ -1,0 +1,102 @@
+---
+KIP: "0014"
+Title: Updated KIP-0012 Account Protocols
+Author: Stuart Popejoy stuart@kadena.io, Doug Beardsley doug@kadena.io
+Status: Draft
+Type: Standard
+Category: Chainweb
+Created: 2021-12-22
+---
+
+## Abstract
+
+Augment KIP-0012 account protocols with two new protocols, Guard Accounts and Rotatable Accounts.
+Modify KIP-0012 Single-Key Accounts to forbid rotation.
+
+## Motivation
+
+KIP-0012 introduced Account Protocols and the Single-Account protocol,
+as implemented in the KDA `coin` contract and elsewhere.
+KIP-0012 reserves all account names starting with `C:` where `C` is a single alpha protocol identifier.
+
+The KIP-0012 Single-Account protocol specifies `k:KEY` where KEY is a hex lowercase ED25519 public key.
+This functions to _reserve the account name_ on any chain which avoids problems like
+"squatting" where an account for a given name on one chain is grabbed by a malicious actor on
+another chain. `k:` has effectively neutralized this threat vector.
+
+However, `k:` is deficient in the following ways:
+
+### Rotation considered harmful for "k:" accounts
+
+Under `fungible-v2` and other standards, `rotate` is specified where an account guard can be
+changed. This presents a challenge to many systems that are accustomed to single-key standards
+on other blockchains such as Bitcoin and Ethereum where an "address" is irrevocably tied
+to a single key. As such this presents problems for "total chain analysis" of a given token
+using standard methods like tracking `fungible-v2.TRANSFER` events, which only provide account
+names and not guards.
+
+While rotation has proven useful with `k:` accounts, this is largely because `k:` makes _account
+reservation_ possible, but by allowing rotate, this robs the protocol from a significant benefit
+which is simplified and robust chain accounting.
+
+### Rotatable Accounts
+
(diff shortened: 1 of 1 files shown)


ZodiacWizard:
Explain in details

emilypi:
See: [rendered](https://github.com/kadena-io/KIPs/blob/d15bdc06468c6687d8778b93e2fcac1eb766a28e/kip-0014/kip-0014.md)

sirlensalot:
> See: [rendered](https://github.com/kadena-io/KIPs/blob/d15bdc06468c6687d8778b93e2fcac1eb766a28e/kip-0014/kip-0014.md)

This will break if the md gets edited, I put a branch-based link in the description https://github.com/kadena-io/KIPs/blob/kip-0014/kip-0014/kip-0014.md

CryptoPascal31:
Maybe, should you take advantage of this new KIP to add some other protocols identifier:

- m: XXXX where XXXX is a module name. It enforces the usage of corresponding module-guard for the account
- n: XXXX:pubKey where XXXX is any string. Same as k; account, but let the user to add for a chosen name for its account. For example n:Pascal:38e5fd698..... I don't know if this is possible in PACT, because of the lack of a split function.
- Other ideas ....

DanCathers:
I am new to crypto and blockchain.  I can't say that I follow your proposal.  What I can say is that I am disappointed with the experimentation I have done so far.  I got coins from the Testnet faucet and tried to use PactyParrots dApp.  I was told that I did not have a balance on Chain 0.  Sure enough, the faucet deposited coins on Chain 1.  I tried to initiate a cross-chain transfer, but I have to have coins on the other chain before I can have coins on the other chain.  Next I tried to just create an account on Chain 0.  I thought I figured out how to do that, but even though the transition said it was successful and charged a gas fee, there is no account on Chain 0.  Just for the hell of it, I tried to have the faucet create a new account using the k: prefix for the name.  It did not like that.

Overall, I can't imagine how anyone who is not a programmer could use your blockchain.  I have programming experience and have not had much luck with your sample apps.  What I want is to buy coins and be certain that they are not going to wind up somewhere that someone on my account name.  I don't care what chain the coins are on and I don't care what chain the apps are on.  I just want it to work fluidly and effortlessly with all the chains under the hood and invisible.  I have thought about trying to buy KDA and hold it in Zelcore, but I don't get the reassuring sense that they know how k: prefix and now it appears that you want to change how that works (not that I have any idea how to use the current functionality).
:-(

kip-0014/kip-0014.md
```diff
@@ -0,0 +1,102 @@
+---
+KIP: "0014"
+Title: Updated KIP-0012 Account Protocols
+Author: Stuart Popejoy stuart@kadena.io, Doug Beardsley doug@kadena.io
```
mohammadfarari1360, line 4: Okay

kip-0014/kip-0014.md
```diff
@@ -0,0 +1,102 @@
...
+---
+
+## Abstract
+
+Augment KIP-0012 account protocols with two new protocols, Guard Accounts and Rotatable Accounts.
+Modify KIP-0012 Single-Key Accounts to forbid rotation.
+
+## Motivation
+
+KIP-0012 introduced Account Protocols and the Single-Account protocol,
```
mohammadfarari1360, line 18: Kip_0012

mohammadfarari1360, line 18: Kip_0012

kip-0014/kip-0014.md
mohammadfarari1360, line 4: Author

kip-0014/kip-0014.md
```diff
@@ -0,0 +1,102 @@
...
+Under `fungible-v2` and other standards, `rotate` is specified where an account guard can be
+changed. This presents a challenge to many systems that are accustomed to single-key standards
+on other blockchains such as Bitcoin and Ethereum where an "address" is irrevocably tied
+to a single key. As such this presents problems for "total chain analysis" of a given token
+using standard methods like tracking `fungible-v2.TRANSFER` events, which only provide account
+names and not guards.
+
+While rotation has proven useful with `k:` accounts, this is largely because `k:` makes _account
+reservation_ possible, but by allowing rotate, this robs the protocol from a significant benefit
+which is simplified and robust chain accounting.
```
pax-k, line 40: Any updates on KIP14? Asking because I'm trying to implement fraud analytics on Kadena and as mentioned here, theoretically a bad actor could rotate a k:account's key, making tracing harder.

PR #29, opened 2022-09-30
KIP-0015: quicksign signing api v1

jmininger:


kip-0015.md
```diff
@@ -0,0 +1,304 @@
...
+- The `SigData` type includes a `hash` field [https://github.com/kadena-io/pact/blob/957b8bd7644cebc60d043b33f0d7ffc53c65d783/src/Pact/Types/SigData.hs#L62-L66](https://github.com/kadena-io/pact/blob/957b8bd7644cebc60d043b33f0d7ffc53c65d783/src/Pact/Types/SigData.hs#L62-L66) . This data can also be derived from the `cmd` field of `SigData`. Imagine a scenario where a malicious dapp sends a `SigData` where the hash represents a malicious transaction, while the payload represents an unrelated transaction that the user is expecting. If the wallet does not explicitly check the hash, by recalculating, this becomes a security issue. To avoid this we force the wallet to compute the hash (by omitting it from the new `CommandSigData` structure)
+
+# Specification
+
+## Schema Sketch
+
+### Request
+```
+{
+  "commandSigDatas": [ { "cmd": string, "signatureList": [ { "publicKey":string, "signature": string|null }]}]
```
mightybyte, line 34: @jmininger After discussing this with @alber70g, we've concluded that we should try to match the format of Pact's `Command` type as closely as possible.  So that would suggest the following changes:

```
{
  "cmds": [ { "cmd": string, "sigs": [ { "pubKey":string|null, "sig": string|null }]}]
}
```

So we're suggesting the following changes:

1. `commandSigDatas` becomes `cmds` to match https://api.chainweb.com/openapi/pact.html#tag/endpoint-send/paths/~1send/post
2. `signatureList` becomes `sigs`, matching https://api.chainweb.com/openapi/pact.html#tag/model-command
3. `signature` becomes `sig` for the same reason
4. `publicKey` becomes `pubKey` and also becomes optional, which allows this API to work transparently with `Command`.

I think `CommandSigData` is still a fine name.  It is simply `Command` with the ability to optionally specify `pubKey` at the top level to make signing simpler.

mightybyte (changes requested):

kip-0015.md
Luzzotica: I like these requested changes.
One of my biggest qualms with the current signing API is that I can't just send a command to the wallet to be signed, the data has been flattened, and some of the names have been changed. This is also very true in the Pact-lang-api... but that's a conversation for another time.

Luzzotica:
I would like to suggest swapping `QsError` to `QuicksignError`.
I like keeping APIs very declarative in their types. Might be semantics, but when i first read QsError I had to think about it for a second before I knew exactly what it was referring to, which seems stupid now, but not at the time. =)

kip-0015.md
```diff
@@ -0,0 +1,304 @@
...
+{
+  "responses": [Response]
+}
+```
+
+or
+
+```
+{
+  "error": QsError
```
Luzzotica, line 108: I would like to suggest swapping QsError to QuicksignError.
I like keeping APIs very declarative in their types. Might be semantics, but when i first read QsError I had to think about it for a second before I knew exactly what it was referring to, which seems stupid now, but not at the time. =)

kip-0015.md
jmininger: My thoughts on each are as follows:
Re:
1) They aren't `cmds`. They are capable of representing states that a `cmd` is not capable of representing. They also dont carry the hash like a `cmd` does. Maybe, just maybe, I could be convinced that `payloads` is a better name since it's more accurate
2) Again, it's not the same as `sigs` in `Command` though, bc it allows optional sigs, which is something the Command type does not. 
3) Sounds good
4) Changing publicKey to `pubKey` sounds good to me. Allowing it to be optional on the other hand allows for the following:
```
[ { "sig": "<sig-here>"}, {}, { "pubKey": <key>, "sig": "<sig here>" } ]
```
This doesn't feel like it gets us any closer to compat with `Command`, and removing the pubkey makes this way less ergonomic for consumers of the api.

kip-0015.md
jmininger: Ah interesting. Note that `QsError` is not actually a field/key you will see in the api when using it. It is just a non-terminal symbol used to specify this (psuedo)-grammar. That being said, I wrote it to mirror the haskell implementation more, and perhaps it would make sense to switch some of the symbol's name to better the document's readability. Thanks for the suggestion!

Luzzotica: I disagree. 

If I'm using TypeScript or any typed language I will create a QuicksignError type that can load the JSON of the response.

And even if it is in an untyped language like Javascript or Python, you will still want to clearly define the types you might receive and give them name so developers can know precisely what to expect. 

Thus, every time you interact with Quicksign and you get an error, you must know what a QuicksignError is. 

Be rigorous and consistent in your definitions and people will like building with your stuff that much more.

kip-0015.md
Luzzotica: Ok, I see what you're saying. I would suggest something like this then:
`CommandSigData` -> `CommandSignatureRequest`
`cmd` -> `commandPayload`
`signatureList` -> `signatureRequestList` or `requiredSigList`

Don't care too much about optional on the pubKey, but the result from my above info would be this:

```json
{
  "commandSignatureRequests": [ { "commandPayload": string, "signatureRequestList": [ { "pubKey":string|null, "sig": string|null }]}]
}
```

kip-0015.md
jmininger: Changed

kip-0015.md
```diff
@@ -0,0 +1,304 @@
...
+
+- The `pubKey` field is a lowercase base-16 encoded public key
+- The `sig` element is either a lowercase base-16 encoded signature or a null value (representing a request for signature).
+
+```
+{
+  "pubKey": string,
+  "sig": string //optional; its omission represents a signature that has not yet been acquired
+}
+```
```
alber70g, line 95: I would suggest the following
- `pubKey` is optional (i.e. can be `undefined`), but cannot be `null`
- `sig` can be either `null` or `undefined`

The reasoning is two-fold:
1. from the request perspective, there is no reason to include the pubKey at this position _however_
2. as is it is part of the response, it improves the user experience to have it optional. This allows the client to accumulate signatures over multiple requests, without having to strip out public keys

```suggestion
## Signer

- The `pubKey` field is a lowercase base-16 encoded public key
- The `sig` element is either a lowercase base-16 encoded signature or a null value (representing a request for signature).

```
{
  "pubKey": string | undefined, // optional; can not be `null`
  "sig": string | null | undefined // optional; its omission represents a signature that has not yet been acquired
}
```

> `undefined` meaning the key is not there
> `null` meaning `"sig": null`
```

kip-0015.md
```diff
@@ -0,0 +1,304 @@
...
+
+- The `cmds` field contains the list of requests; some of which contain new signatures attached. Even in the case of an error, it is still useful to have the ordered list of `CommandSigDatas` in the response, given that some of them will contain new signatures.
+
+```
+{
+  "commandSigData": CommandSigData,
+  "outcome": Outcome
+}
+```
+
```
alber70g, line 122: CommandSigData from the request, is not the same as it is from the response. Fields are not optional in the response that are in the request.

- `sigs` will not contain nulls, as there will always be a `{ "pubKey": string }`
- so `Signer`-type will have `"pubKey": string`
- `sig` will be `null` to make explicit that there's a signature missing

```suggestion
## CommandSigDataResponse

```ts
type CommandSigDataResponse = {
  "sigs": [SignerResponse],
  "cmd": string
}

type SignerResponse = {
  "pubKey": string,
  "sig": string | null | undefined,
}
```

```

mightybyte (approved):

PR #30, opened 2022-10-19
KIP-0016: URI Scheme

tim-eucalyptus:
This KIP proposes a URI scheme for making Kadena payments.

## Motivation

The convenience of representing payment requests by standard URLs has been a major factor in the wide adoption of Bitcoin, Ethereum and others. Bringing a similarly convenient mechanism to Kadena would speed up its acceptance as a payment platform among end-users. In particular, URIs encoded as QR codes ease the process of requesting and sending payments via smart phones and end user applications. Thus, payment request URIs could be delivered through a very broad, ever growing selection of channels and accelerate the adoption and usablity of Kadena.

### DIFF
--- a/kip-0016.md
+++ b/kip-0016.md
@@ -1,0 +1,98 @@
+---
+KIP: 0016
+Title: URI Scheme
+Author: Tim Molter tim@eucalyptuslabs.com
+Status: Draft
+Type: Standard
+Category: Applications
+Created: 2022-10-19
+---
+
+## Abstract
+
+This KIP proposes a URI scheme for making Kadena payments.
+
+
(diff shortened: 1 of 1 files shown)


mightybyte:
Thanks for the proposal!  I like the direction.  One question to clarify what you're proposing...what do you mean by the syntax `[Number]`?  Do the brackets mean mean that it's optional?  That it's a list?  Something else?

tim-eucalyptus:
I will clarify that and also make a note about what is optional or not.

tim-eucalyptus:
@mightybyte I've pushed some minor clarifications.

PR #31, opened 2022-11-15
KIP-0017: wallet-connect-v2 sign-api session spec

jmininger:
Update March 8, 2023:
We've updated the Wallet Connect Sign API v2 specifications for Kadena. Some of the major changes we introduced:
- Re-organization and re-wording of some section.
- Added specifications for a `getAccounts` method for retrieving Kadena account information from a public key.
- Made the method names versioned to allow for future upgradability.

We are aiming to finalize this KIP by the end of this week, so Friday, March 10, 2023.

### DIFF
--- a/kip-0017.md
+++ b/kip-0017.md
@@ -14,17 +14,17 @@ Created: 2022-11-15
     - [Example Pairing Proposal Request and Settlement Response](#example-pairing-proposal-request-and-settlement-response)
     - [WalletConnect Proposal Request](#walletconnect-proposal-request)
     - [WalletConnect Settlement Response](#walletconnect-settlement-response)
-    - [Wallet Connect Chain IDs](#wallet-connect-chain-ids)
-    - [Wallet Connect Accounts](#wallet-connect-accounts)
+    - [WalletConnect Chain IDs](#wallet-connect-chain-ids)
+    - [WalletConnect Accounts](#wallet-connect-accounts)
     - [WalletConnect Methods](#walletconnect-methods)
       - [kadena_getAccounts_v1](#kadena_getaccounts_v1)
         - [Example kadena_getAccounts_v1 Request and Response](#example-kadena_getaccounts_v1-request-and-response)
         - [kadena_getAccounts_v1 Request](#kadena_getaccounts_v1-request)
         - [kadena_getAccounts_v1 Response](#kadena_getaccounts_v1-response)
-      - [kadena_quickSign_v1](#kadena_quicksign_v1)
-        - [Example kadena_quickSign_v1 Request and Response](#example-kadena_quicksign_v1-request-and-response)
-        - [kadena_quickSign_v1 Request](#kadena_quicksign_v1-request)
-        - [kadena_quickSign_v1 Response](#kadena_quicksign_v1-response)
+      - [kadena_quicksign_v1](#kadena_quicksign_v1)
+        - [Example kadena_quicksign_v1 Request and Response](#example-kadena_quicksign_v1-request-and-response)
+        - [kadena_quicksign_v1 Request](#kadena_quicksign_v1-request)
+        - [kadena_quicksign_v1 Response](#kadena_quicksign_v1-response)
       - [kadena_sign_v1](#kadena_sign_v1)
         - [Example kadena_sign_v1 Request and Response](#example-kadena_sign_v1-request-and-response)
         - [kadena_sign_v1 Request](#kadena_sign_v1-request)
@@ -38,7 +38,7 @@ Created: 2022-11-15
 # Motivation
 
 This KIP proposes that the Kadena ecosystem implement the
-[Wallet Connect v2 Sign API](https://docs.walletconnect.com/2.0/api/sign). This
+[WalletConnect v2 Sign API](https://docs.walletconnect.com/2.0/api/sign). This
 API allows for establishing a secure channel for communicating between dApps and
 wallets in a platform-agnostic manner.
 
@@ -129,14 +129,14 @@ The response for the Proposal contains a `namespaces` object with the following 
     1.2. `methods`: `Array<string>` - an array of [Methods](#walletconnect-methods) available.\
     1.3. `events`: `Array<string>` - an array of [Events](#walletconnect-events) available.
 
-## Wallet Connect Chain IDs
+## WalletConnect Chain IDs
 
-Chain IDs are Wallet Connect’s way of identifying specific blockchains within an ecosystem.
+Chain IDs are WalletConnect’s way of identifying specific blockchains within an ecosystem.
 
 They are defined by the [CAIP-2 standard](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md) and are formatted as:
 
 ```
-    <namespace> + ":" + <reference>
+<namespace> + ":" + <reference>
 ```
 - `<namespace>`: `string` - always defined as "kadena".
 - `<reference>`: `string` - a unique network ID within the blockchain ecosystem (e.g. `testnet04`, `mainnet01`).
@@ -153,19 +153,19 @@ These are valid Chain IDs:
 > We are displaying Kadena `chainId` information in the [`kadena_getAccounts_v1`](#kadena_getAccounts_v1) method.
 
 
-## Wallet Connect Accounts
+## WalletConnect Accounts
 
-Accounts are Wallet Connect's way of specifying control over signing privileges.
+Accounts are WalletConnect's way of specifying control over signing privileges.
 
 They are defined by the [CAIP-10 standard](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md), and are formatted as:
 ```
-    <chain_id> + ":" + <account_address>
+<chain_id> + ":" + <account_address>
 ```
-- `<chain_id>`: `string` - the Wallet Connect [Chain ID](#wallet-connect-chain-ids).
+- `<chain_id>`: `string` - the WalletConnect [Chain ID](#wallet-connect-chain-ids).
 - `<account_address>`: `string` - the public key available for signing by the wallet.
 
 ```jsonc
-// Wallet Connect Account Example
+// WalletConnect Account Example
 
 "kadena:mainnet01:38298612cc2d5e841a232bd08413aa5304f9ef3251575ee182345abc3807dd89"
 ```
@@ -180,7 +180,7 @@ The use of account names that can be derived from the public key (i.e. `k:`-acco
 
 ## WalletConnect Methods
 
-Wallet Connect methods are ways for dApps to interact with a wallet using predefined method identifiers and an optional payload.
+WalletConnect methods are ways for dApps to interact with a wallet using predefined method identifiers and an optional payload.
 
 The methods available for Kadena are:
 
@@ -281,7 +281,7 @@ This method also expects a `params` object with the following properties:
 
 1. `params`: `Object`\
     1.1. `accounts`: `Array<Object>`\
-    __ 1.1.1 `account`: `string` - a [Wallet Connect Account](#wallet-connect-accounts) for which the Kadena account names are requested.\
+    __ 1.1.1 `account`: `string` - a [WalletConnect Account](#wallet-connect-accounts) for which the Kadena account names are requested.\
     __ 1.1.2 `contracts`: `Array<string>` (optional) - array of contracts for which the Kadena account name is requested. Returns all known Kadena accounts when omitted.
 
 > **NB**: We choose to not add chains to the [getAccounts_v1 request](#kadena_getAccounts_v1-Request) since it would
@@ -293,8 +293,8 @@ The method response expects a `result` object with the following properties:
 
 1. `result`: `Object`\
     1.1. `accounts`: `Array<Object>`\
-    __ 1.1.1 `account`: `string` - the requested [Wallet Connect Account](#wallet-connect-accounts).\
-    __ 1.1.2 `publicKey`: `string` - the requested public key, which was extracted from the Wallet Connect account.\
+    __ 1.1.1 `account`: `string` - the requested [WalletConnect Account](#wallet-connect-accounts).\
+    __ 1.1.2 `publicKey`: `string` - the requested public key, which was extracted from the WalletConnect account.\
     __ 1.1.3 `kadenaAccounts`: `Array<Object>`\
     ____ 1.1.3.1 `name`: `string` - the account name, as stored in the blockchain, for the specified contract.\
     ____ 1.1.3.2 `contract`: `string` - the contract on which this account name is present.\
@@ -314,14 +314,14 @@ The method response expects a `result` object with the following properties:
 QuickSign is part of the [Kadena Signing API](https://github.com/kadena-io/signing-api) and was defined in [kip-0015 (QuickSign Signing API v1)](https://github.com/kadena-io/KIPs/blob/master/kip-0015.md).
 This method allows the wallet to show the user multiple transactions that need signature approval.
 
-#### Example kadena_quickSign_v1 Request and Response
+#### Example kadena_quicksign_v1 Request and Response
 
 ```jsonc
-// kadena_quickSign_v1 Method Request
+// kadena_quicksign_v1 Method Request
 {
   "id": 1,
   "jsonrpc": "2.0",
-  "method": "kadena_quickSign_v1",
+  "method": "kadena_quicksign_v1",
   "params": {
     "commandSigDatas": [CommandSigData] // type defined in KIP-15
   }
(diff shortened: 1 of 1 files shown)


tim-eucalyptus:
There is a problem here with the way accounts are defined in this KIP. Perhaps it's a typo? It is presented as `chain_id + ":" + account_address`, but then in the example instead of `account_address`, we have what looks like public keys. Sending public keys to the app will not work because the apps cannot assume the public key maps to the account: `k:pub_key'.

Screenshot 2023-01-26 at 2 47 46 PM

tim-eucalyptus:
Furthermore, when defining an account in k-address format, we cannot shouldn't use the ":" in the k-address (k:ajkhsgdjkyadgjadjghajsdfjashdfg) but rather according to the linked to CAIP-10 the URL-encoded version. So the correct way to do it is to simply URL-encode the ":". This would be a way better candidate for a standard as it's blockchain agnostic.
i.e. : ==> %3A I have seen a hack going around where it is replaced with "**". 

For example:

kadena:mainnet01:k%3A22ddc64851718e9d41d98b0f33d5e328ae5bbbbd97aed9885adac0f2d070ff9c

Screenshot 2023-01-26 at 2 40 48 PM

tim-eucalyptus:
And finally, can we get some closure on this, as in get this accepted? There are many wallets and apps now implementing WCv2 and the chance for all of us to follow some standard, may slip away.

jmininger:
Nope, not a mistake. Wallet connect's `accounts` actually map more cleanly to pact's `signers` (public keys), because it is derived from the ethereum model of `accounts == pubkey`. Kadena "accounts" are not pact natives, and their implementation can vary at the contract level. Consider that accounts don't actually make sense without the context of a specific contract that they are being stored in.  In many cases, there are contracts that you want interact with (sign for), that don't necessarily involve an account. 

As for dapps getting access to the account: I can see how there might be specific cases where this can be useful. Here are some possibilities:

1) Continue to have it be a manual operation where the user specifies their account to the dapp
2) Assume it is a k-account (I honestly think this is sufficient for 99% of cases and can be used in combination with 1)
3) Have the dapp look up all associated accounts using an indexer. 
4) Add an additional wallet-connect method (in addition to sign/quicksign) that requests accounts + keyset + contract from the wallet. Though you need to consider that not all kadena wallets have a notion of "accounts" since they are not natives (ex: https://github.com/kadena-community/keychain )

tim-eucalyptus:
I appreciate your answer. There's still confusion in naming it `account_address` IMO. It should be named what it is: the `pubKey`, or at least explained in the text better. 

I still believe having the wallet send the k-address or vanity account associated with an underlying pubkey is much better. That way is in line with the spirit of wallet connect where the account or address is what is communicated. Unfortunately for Kadena since the account/address is not one-to-one mapped to the pubkey, more wallet connect calls are needed to associate pub keys with addresses. So I would do it the other way - first address then pub keys. Otherwise for the 1% where the k-address cannot be assumed it fits to the pubkey, it's going to fail. 

Most importantly, some standard should be decided upon soon, otherwise each wallet and dapp will be building their own standards like what happened with Kadena NFTs.

jmininger:
> There's still confusion in naming it account_address IMO. It should be named what it is: the pubKey, or at least explained in the text better.

Well `account_address` is per the wallet-connect standard so we can't do much to change that. In anticipation of this confusion, I added the all-caps + bold *NOTE* section that you screenshot, but evidently it was not clear enough :laughing: . Any suggestions to improve that section are welcome. 

> I still believe having the wallet send the k-address or vanity account associated with an underlying pubkey is much better. That way is in line with the spirit of wallet connect where the account or address is what is communicated.

I strongly disagree. Wallet-connect is a protocol that facilitates access to a wallet's most valuable asset -- its keypair. Communication needs to be at that level, not at the level of contract state. Furthermore, it also makes sense because both the sign + quicksign methods are requests for _signatures_. The way these work is that the dapp needs to construct requests around specific keys (pact calls them "signers"). Again, some wallets in the kadena ecosystem do not even operate at the account level, since it is a contract-based piece of state. 

> Otherwise for the 1% where the k-address cannot be assumed it fits to the pubkey, it's going to fail.
Not if you provide an option for the user to change the address at the dapp level. 
If you really want to add an additional method that gets accounts, what i recommend is the following: 
https://docs.walletconnect.com/2.0/specs/clients/sign/session-namespaces should provide a way for the wallet to allow for additional permissions not requested by the dapp. You can add your own custom method that provides account names, and dapps can chose whether or not to use this feature. If it's a successful experiment, I dont see why kadena wouldn't accept a new KIP to add it. That being said, I don't see it as a critical feature bc of the reasons and possible solutions given above. 

> Most importantly, some standard should be decided upon soon, otherwise each wallet and dapp will be building their own standards like what happened with Kadena NFTs.
Agreed. Gonna tag @mightybyte @Randynamic @alber70g since this is something i no longer actively work on.

tim-eucalyptus:
OK, I did read that NOTE, but to me it wasn't clear enough I suppose. Oh well.

Just to throw out a counter example to your argument about passing public keys, I'd like to point out Hedera. According to the official CAIP 10 (https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md), which your KIP 17 references, we can see that in their test cases:

```
# Hedera address (with optional checksum suffix per [HIP-15][])
hedera:mainnet:0.0.1234567890-zbhlt
```

Here the "address" and not the public key is used. So if it was decided to call public keys "addresses" or "accounts" for Kadena, this would make it an odd ball.

Additionally, nowhere on the CAIP does it mention public keys.

In many cases, if a k-address is sent the dapp can assume the pubkey and generate the TX to be signed without any other information and k-addresses are the modern and "safe" way to go in Kadena as far as my understanding goes. In order to be backwards compatible with old squattable addresses WC will need a secondary call you get the pubkey(s) in order to build the TX for the wallet to sign.

How would your way handle multi-sig looking forward? It seems to me it would get pretty sloppy. First you get the first pubkey (where I and CAIP are proposing sending the account) through the standard channel, then you need another call to get additional pub keys and another one to get the account.

Anyway, I'm not trying to argue here publicly, and it's 100% your decision. I'm just giving my two cents about how I see it from my admittedly non-expert perspective.

tim-eucalyptus:
Perhaps a `getKeySet` or `getAccountGuard` would be the appropriate additionally-needed call you're referring to. That would fit nice. We could stick to the CAIP-0010 standard and then get the underlying key(s) via a second call if needed. This works for the old kadena arbitrary addresses and also the new k-address format. It would also support multi-sig I believe.

Luzzotica:
I would like to support @tim-eucalyptus here!

Sending the account makes it more flexible. Allows you to send multisig accounts, vanity accounts, etc. 

And if you REALLY need the public key, the dApp can request that themselves.

jermaine150:
@mightybyte @tim-eucalyptus @Luzzotica 
We've updated the specification and added rationale behind the choices that were made. Please let us know if you have anymore feedback, we're aiming to finalise this KIP by the end of this week.

Randynamic (dismissed):

kip-0017.md
```diff
@@ -0,0 +1,489 @@
...
+
+
+## SLIP-0044
+
+While not required in the namespace proposal process, a network's SLIP-0044
+value is used throughout the WalletConnect demo code.
+
+Kadena's coin type is
+**626** as documented in
+https://github.com/satoshilabs/slips/blob/master/slip-0044.md.
```
jwiegley, line 483:
```suggestion
(SLIP-0044)[https://github.com/satoshilabs/slips/blob/master/slip-0044.md].
```

kip-0017.md
```diff
@@ -0,0 +1,489 @@
...
+API allows for establishing a secure channel for communicating between dApps and
+wallets in a platform-agnostic manner.
+
+This is beneficial to Kadena for the following reasons:
+
+- The Kadena ecosystem doesn't have a standard protocol for wallets and dApps to
+  communicate with each other.
+- The Wallet Connect (WC) protocol is tried and
+  tested across all devices and it removes the burden of having to maintain a
+  kadena-specific implementation.
```
jwiegley, line 51:
```suggestion
  Kadena-specific implementation.
```

kip-0017.md
```diff
@@ -0,0 +1,489 @@
...
+
+These are valid Chain IDs:
+
+- `kadena:mainnet01` - Kadena mainnet
+- `kadena:testnet04` - Current Kadena testnet
+- `kadena:development` - Kadena devnet
+
+> **NB**: `Chain IDs`, as defined here, are semantically specific to
+> WalletConnect and are not equivalent to a Kadena "Chain ID" (e.g. chain 1).
+> They are more closely related to Kadena's `networkId` field.
```
jwiegley, line 152: Why aren't they called "Network IDs", then?

kip-0017.md
```diff
@@ -0,0 +1,489 @@
...
+> We are displaying Kadena `chainId` information in the [`kadena_getAccounts_v1`](#kadena_getAccounts_v1) method.
+
+
+## Wallet Connect Accounts
+
+Accounts are Wallet Connect's way of specifying control over signing priviledges.
```
jwiegley, line 158:
```suggestion
Accounts are Wallet Connect's way of specifying control over signing privileges.
```

kip-0017.md
```diff
@@ -0,0 +1,489 @@
...
+This is subject to change in the future as we understand the workflows more through extended use of the protocol.
+
+However, any future Kadena event names will be formatted as:
+```
+    kadena_<eventName>_v<eventVersionNumber>
+```
+- `<eventName>`: `string` - name of the event in camelCase format.
+- `<eventVersionNumber>`: `integer` - the version number of the event.
+
+> **NB**: All event names will be versioned to allow for facilitate future
```
jwiegley, line 445:
```suggestion
> **NB**: All event names will be versioned to allow for future
```

kip-0017.md
```diff
@@ -0,0 +1,489 @@
...
+Several options were considered when discussing what information to include
+in the `accounts` section of the initial pairing response.
+
+However, the following reasons forced us to only include the public keys
+(prefixed with Kadena's Wallet Connect Chain ID):
+
+1. Kadena fungible contracts (specifically the `coin` contract) have a maximum account name length of 256, but
+   the [CAIP-10 standard](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md)
+   limits wallet connect account names to a maximum of 128 characters.
+2. We considered only supporting `k:` accounts, as they're build from
```
jwiegley, line 462:
```suggestion
2. We considered only supporting `k:` accounts, as they're built from
```

kip-0017.md
```diff
@@ -0,0 +1,489 @@
...
+   `k`:`publicKey`, but we want this response to be generic for all accounts
+   that could exist on the Kadena blockchain.
+
+Therefore, we **strongly** advice builders to integrate the [`kadena_getAccounts_v1`](#kadena_getAccounts_v1)
```
jwiegley, line 466:
```suggestion
Therefore, we **strongly** advise builders to integrate the [`kadena_getAccounts_v1`](#kadena_getAccounts_v1)
```

kip-0017.md
```diff
@@ -0,0 +1,489 @@
...
+method into the pairing process with the wallet.
+
+This will allow them to support multi-sig accounts (e.g. `w:`-accounts) and
+other means of singing transactions.
```
jwiegley, line 470:
```suggestion
other means of signing transactions.
```

kip-0017.md
jermaine150: In this context the name "Chain ID" is defined by the CAIP-2 standard which the Wallet Connect specification adheres to. We're unable to change that name, but merely added this note to avoid confusion with Kadena's Chain ID.

alber70g: We want to use the naming that WalletConnect uses in places where it's dictated by WalletConnect

alber70g:
@tim-eucalyptus 
> Here the "address" and not the public key is used. So if it was decided to call public keys "addresses" or "accounts" for Kadena, this would make it an odd ball.

The rationale is that we can't map "Kadena Accounts" to [caip-10 account_address](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md#syntax). The CAIP-10 states that an account address can be `account_address:   [-.%a-zA-Z0-9]{1,128}`. This is not compatible with "Kadena Account Names" as they can be 256 chars long.

alber70g (changes requested):

kip-0017.md
```diff
@@ -0,0 +1,489 @@
...
+method can be called.
+
+## WalletConnect Methods
+
+Wallet Connect methods are ways for dApps to interact with a wallet using predefined method identifiers and an optional payload.
+
+The methods available for Kadena are:
+
+- [`kadena_getAccounts_v1`](#kadena_getAccounts_v1)
+- [`kadena_quickSign_v1`](#kadena_quicksign_v1)
```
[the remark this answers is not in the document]
LindaOrtega: Yes, there are a couple of places that we use the `kadena_quickSign_v` string literal, so we have to change it in all those places.

kip-0017.md
```diff
@@ -0,0 +1,489 @@
...
+
+> **NB**: All method names will be versioned to allow for future
+> upgradability of the Wallet Connect method APIs.
+
+### kadena_getAccounts_v1
+
+This method returns the fungible account names associated with a specific Wallet Connect account (i.e. a prefixed public key) in the Kadena blockchain.
+
+These Wallet Connect accounts are retrieved from the `accounts` property of a [Paring response](#example-pairing-proposal-request-and-settlement-response).
+
```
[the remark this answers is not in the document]
LindaOrtega: The `(i.e. a prefixed public key)` was an attempt to quickly describe what we are returning in WalletConnect accounts. WalletConnect accounts in the Kadena implementation are public keys with a prefix (`kadena:testnet04:` for example). 

In this section, there are two types of accounts we are using, kadena account and WalletConnect accounts, so I wanted to add some additional descriptions.

kip-0017.md
```diff
@@ -0,0 +1,489 @@
...
+> **NB**: `accounts` is semantically specific to WalletConnect. This does not
+> represent an account in a Kadena fungible contract (e.g. `coin`).
+> Read the motivation for this in [Accounts vs. Public Keys](#accounts-vs-public-keys).
+
+Some users will want to know which account names in a Kadena fungible contract are associated with the public keys returned by Wallet Connect.
+
+These users have two options:
+- Some account names can be derived from the public key (e.g. `k:` accounts that are defined as `k:<public_key>`).
+- For all other cases, the [`kadena_getAccounts_v1`](#kadena_getAccounts_v1)
+method can be called.
```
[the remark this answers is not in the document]
LindaOrtega: Adding "Kadena" in front of accounts for some clarity.

```suggestion
After the initial pairing with the wallet via WalletConnect, the Kadena accounts relevant for each public key can be requested by calling [getAccounts](#kadena_getaccounts_v1).

The use of account names that can be derived from the public key (i.e. `k:`-accounts that are defined as `k:<public_key>`), is **highly discouraged** as implementing this will limit the end-users flexibility in which account to use, thus making dApps less generic.
```

kip-0017.md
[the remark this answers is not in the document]
alber70g: Ah yes, maybe an example would be more describing?

kip-0017.md
```diff
@@ -0,0 +1,489 @@
...
+
+The methods available for Kadena are:
+
+- [`kadena_getAccounts_v1`](#kadena_getAccounts_v1)
+- [`kadena_quickSign_v1`](#kadena_quicksign_v1)
+- [`kadena_sign_v1`](#kadena_sign_v1)
+
+Any additional Kadena method names will be formatted as:
+```
+    kadena_<methodName>_v<methodVersionNumber>
```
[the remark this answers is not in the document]
LindaOrtega: There are other places that add this tab

kip-0017.md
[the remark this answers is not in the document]
LindaOrtega: I edited the suggestion to use `Kadena account name` to help clarify which account we're referring to.

Randynamic (dismissed):

jwiegley (dismissed):

alber70g (approved):

jermaine150 (approved):

tim-eucalyptus:
In order for Kadena to be listed on the Wallet Connect [chain list](https://docs.walletconnect.com/2.0/advanced/multichain/chain-list) (and allow for deep linking to work) , it's necessary to set the [CASA - chain-agnostic namespace](https://github.com/ChainAgnostic/namespaces)

On Jan 26th, I opened an issue with them here: https://github.com/WalletConnect/walletconnect-monorepo/issues/1905

But to continue, someone from Kadena needs to submit a PR, adding Kadena here: https://github.com/ChainAgnostic/namespaces

Who is responsible for WalletConnect at Kadena?

LindaOrtega:
Hi @tim-eucalyptus, thank you for bringing this to our attention and for opening up that issue. I've added the work that needs to be done to get us listed on CASA to our backlog for next Sprint.

PR #32, opened 2023-02-22
ag/wallet connect kip 0017

alber70g:
- kip-0017
- chore: base for WalletConnect specification

---
  - https://app.asana.com/0/0/1203959391010080

### DIFF
--- a/kip-0017.md
+++ b/kip-0017.md
@@ -313,57 +313,94 @@ The method response contains a `result` object that contains the following prope
 
 ### kadena_quickSign_v1
 
+TODO Description of quicksign.
+
 #### Example kadena_quickSign_v1 Request and Response
 
 ```jsonc
-// Signing Request
+// kadena_quickSign_v1 Method Request
 {
-  "commandSigDatas": [CommandSigData]
+  "id": 1,
+  "jsonrpc": "2.0",
+  "method": "kadena_quickSign_v1",
+  "params": {
+    "commandSigDatas": [CommandSigData] // type defined in KIP-15
+  }
 }
-```
-
-The structure of `CommandSigData` is defined in the
-[kip-0015 quicksign signing api v1](https://github.com/kadena-io/KIPs/blob/master/kip-0015.md#commandsigdata)
+````
 
 ```jsonc
-// Succesfull Signing Response
+// Successful kadena_quickSign_v1 Response
 {
-  "responses": [Response]
+    "id": 1,
+    "jsonrpc": "2.0",
+    "result":  {
+      "responses": [Response] // type defined in KIP-15
+    }
 }
 ```
 
 or
 
 ```jsonc
-// Failed Signing Response
+// Failed kadena_quickSign_v1 Response
 {
-  "error": QuicksignError
+    "id": 1,
+    "jsonrpc": "2.0",
+    "result":  {
+      "error": QuicksignError // type defined in KIP-15
+    }
 }
 
 ```
 
-The structure of `Response` and `QuicksignError` are defined in the
-[kip-0015 quicksign signing api v1](https://github.com/kadena-io/KIPs/blob/master/kip-0015.md#quicksign-response)
+#### kadena_quickSign_v1 Request
+
+This method needs to be called with a `params` object with the following properties:
+
+1. `params`: `Object`\
+    1.1. `commandSigDatas`: `Array<Object>` - A list of `CommandSigData` to sign. The properties of this type are defined in
+[kip-0015 QuickSign Signing API v1](https://github.com/kadena-io/KIPs/blob/master/kip-0015.md#commandsigdata).
+
+
+#### kadena_quickSign_v1 Response
+
+The method response contains a `result` object that contains the following properties if signing was successful:
+
+1. `result`: `Object`\
+    1.1. `responses`: `Array<Object>` - A list of `Response` contained signed . The properties of this type are defined in
+[kip-0015 QuickSign Signing API v1](https://github.com/kadena-io/KIPs/blob/master/kip-0015.md#response-1).
+
+If signing was NOT successful, then the `result` object contains the following properties:
+1. `result`: `Object`\
+    1.1. `error`: `Object` - A list of `QuicksignError`. The properties of this type are defined in
+[kip-0015 QuickSign Signing API v1](https://github.com/kadena-io/KIPs/blob/master/kip-0015.md#quicksignerror).
 
 ### kadena_sign_v1
+
+TODO Add description for this method.
 
 #### Example kadena_sign_v1 Request and Response
 
 ```jsonc
-// Signing request
+// kadena_sign_v1 Method Request
 {
-  signingCmd
+  "id": 1,
+  "jsonrpc": "2.0",
+  "method": "kadena_sign_v1",
+  "params": {
+    signingCmd // Chainweaver type
+    // TODO: Find actual example
+  }
 }
 
 ```
 
-The structure of `signingCmd` is defined in the
-[Chainweaver Signing Api](https://github.com/kadena-io/pact-lang-api#chainweaver-signing-api-command)
-
 ```jsonc
-# Signing response
+// kadena_sign_v1 Response
 {
   "signedCmd": {
+      // TODO: Find actual example
       cmd: ...,
       hash: ...
       sigs: [
@@ -372,6 +409,19 @@ The structure of `signingCmd` is defined in the
     }
 }
 ```
+
+#### kadena_sign_v1 Request
+
+This method needs to be called with a `params` object with the following properties:
+
+1. `params`: `Object` - A `signingCmd` object whose properties are defined in the
+[Chainweaver Signing Api](https://github.com/kadena-io/pact-lang-api#chainweaver-signing-api-command).
+
+#### kadena_sign_v1 Response
+
+The method response contains a `result` object that contains the following properties:
+
+1. `result`: `Object` - TODO figure out what the properties actually are.
 
 ## WalletConnect Events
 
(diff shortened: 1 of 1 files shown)


kip-0017.md
```diff
@@ -60,53 +111,304 @@ kadena:mainnet01:22ddc64851718e9d41d98b0f33d5e328ae5bbbbd97aed9885adac0f2d070ff9
...
+`accounts` property can be used to retrieve the available KadenaAccounts
+
+#### Example kadena_get_accounts_v1 Request and Response
+
+```jsonc
+// kadena_get_accounts_v1 request
+{
+  "accounts": [
+    "kadena:mainnet01:38298612cc2d5e841a232bd08413aa5304f9ef3251575ee182345abc3807dd89"
+  ]
```
LindaOrtega, line 258: I think these properties are missing from the request example:
```suggestion
       "accounts":[{
        "account": "kadena:mainnet01:38298612cc2d5e841a232bd08413aa5304f9ef3251575ee182345abc3807dd89",
        "contracts": ["coin"]
      }]
```

alber70g: Yes I've added them.

kip-0017.md
```diff
@@ -8,105 +8,426 @@ Category: Interface
...
+These are valid Kadena Chain IDs:
+
+- `kadena:mainnet01` - Kadena mainnet
+- `kadena:testnet04` - Current kadena testnet
+- `kadena:development` - Kadena devnet
+
+> **NB**: `Chain IDs`, as defined here, are semantically specific to
+> WalletConnect and is not equivalent to Kadena's "chain IDs".
+> They are more closely related to Kadena's `networkId` field.
+> <Read the motivation for this in TODO>
```
LindaOrtega, line 161: Added this TODO in case we wanted to document why we aren't including chain ids in the Wallet Connect chain ID.

kip-0017.md
```diff
@@ -8,105 +8,425 @@ Category: Interface
...
+      1.1.1 `account`: `string` - the public key (a Wallet Connect [Account](#wallet-connect-accounts)) for which the Kadena account name is requested.   
+      1.1.2 `contracts`: `Array<string>` (optional) - array of fungible contracts for which the Kadena account name is requested. Returns all known Kadena accounts if omitted.  
+
+#### kadena_getAccounts_v1 Response
 
+The method response contains a `result` object that contains the following properties:
+
+1. `result`: `Object`  
+    1.1. `accounts`: `Array<Object>`  
+    __ 1.1.1 `account`: `string` - the Wallet Connect [Account](#wallet-connect-accounts) requested.
```
LindaOrtega, line 287: The __ are my attempt to get the tabs to work in the preview. Didn't add it to the rest in case we wanted to go with a different approach

kip-0017.md
```diff
@@ -8,105 +8,425 @@ Category: Interface
...
+- For all other cases, the [`kadena_getAccounts_v1`](#kadena_getAccounts_v1)
+method can be called.
+
+## WalletConnect Methods
+
+Wallet Connect methods are ways for dApps to interact with a wallet using predefined method identifiers and an optional payload.
+
+The methods available for Kadena are:
+
+- [`kadena_getAccounts_v1`](#kadena_getAccounts_v1)
```
LindaOrtega, line 192: I camel cased `getAccounts` since it might make it easier to parse

kip-0017.md
```diff
@@ -8,105 +8,425 @@ Category: Interface
...
+// kadena_getAccounts_v1 Method Response
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "result": [
+    {
+      "accounts": [
+        {
+          "account": "kadena:mainnet01:38298612cc2d5e841a232bd08413aa5304f9ef3251575ee182345abc3807dd89",
+          "publicKey": "38298612cc2d5e841a232bd08413aa5304f9ef3251575ee182345abc3807dd89",
```
LindaOrtega, line 231: This used to be `pubKey` so I expanded it to the full `publicKey` for extra clarity

kip-0017.md
```diff
@@ -8,105 +8,425 @@ Category: Interface
...
+  }
+}
+````
+
+```jsonc
+// kadena_getAccounts_v1 Method Response
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "result": [
```
LindaOrtega, line 226: Added the "id", "jsonrpc" fields to getAccounts request and response examples because I couldn't find the Wallet Connect Data Structure used by this Cosmos example when describing their methods: https://docs.walletconnect.com/2.0/advanced/rpc-reference/cosmos-rpc#example

kip-0017.md
```diff
@@ -8,105 +8,426 @@ Category: Interface
...
-# ****Backwards Compatibility****
+## WalletConnect Events
+
+At the moment we see no need to standardize any events and believe that most
+use cases can be handled at the Kadena protocol level that sits atop Wallet
+Connect.
+
+This is subject to change in the future as we understand the workflows more through extended use of the protocol.
+
+However, any future Kadena event names will formatted as:
```
LindaOrtega, line 381: Thought having this section would be useful to make sure any kadena- or community-driven events follow a specific standard.

LindaOrtega: I also added a similar section in the `WalletConnect Methods` section.
Not sure about the wording of the NB section.

jermaine150 (approved):

kip-0017.md
jermaine150: I like this approach, in combination with newlines it looks good in the rendered version now.

LindaOrtega (changes requested):
PR looks good, but the following editing tasks need to be finished:
- Make sure the `quicksign` and `sign` sections match the format we’ve been using in other places
- Editing the table of contents to reflect the updated sections
- Doing a full re-read to look for any errors (paying special attention that the field names in our data types match the examples). 
- I also added the above comments for smaller discussion points.

kip-0017.md
```diff
@@ -8,105 +8,422 @@ Category: Interface
...
+
+> **NB**: The accounts that are returned are the ones that are available in the
+> scope of the wallet (i.e. only the accounts the wallet knows about).
+
+> **NB**: The same Kadena account name can be used in different contracts
+> (e.g. `coin`, `my-fungible`). They require different entries
+> since the chains on which the accounts are present can differ.
+
+> **NB**: We choose to not add chains to the [getAccounts_v1](#kadena_getAccounts_v1-Request) Request, since it would
+> add complexity to the interface. The dApp can filter on the chains needed.
```
LindaOrtega, line 312: @jermaine150 thoughts on moving this section as an NB in the Request section?
This also reminded me that it might be useful to add a sentence in the description of method that states something along the lines of:
"Users will known what contracts and chains contain the account name"

kip-0017.md
```diff
@@ -8,105 +8,422 @@ Category: Interface
...
+```
+
+#### kadena_getAccounts_v1 Request
+
+This method needs to be called with a `params` object with the following properties:
+
+1. `params`: `Object`\
+    1.1. `accounts`: `Array<Object>`\
+    __ 1.1.1 `account`: `string` - the public key (a Wallet Connect [Account](#wallet-connect-accounts)) for which the Kadena account name is requested.\
+    __ 1.1.2 `contracts`: `Array<string>` (optional) - array of fungible contracts for which the Kadena account name is requested. Returns all known Kadena accounts if omitted.
```
LindaOrtega, line 289: I used the word "fungible" in a couple places, but I'm not sure if this should just be omitted since there might be other use cases where you want to retrieve the account name but not its not a fungible contract account name.

LindaOrtega (approved):
I updated the table of contents and finished the styling for the `/quicksign` and `/sign` sections.

kip-0017.md
```diff
@@ -8,105 +8,465 @@ Category: Interface
...
 {
-  "kadena": {
-    "accounts": [
-      "kadena:mainnet01:38298612cc2d5e841a232bd08413aa5304f9ef3251575ee182345abc3807dd89",
-      "kadena:testnet04:38298612cc2d5e841a232bd08413aa5304f9ef3251575ee182345abc3807dd89",
-      "kadena:testnet04:22ddc64851718e9d41d98b0f33d5e328ae5bbbbd97aed9885adac0f2d070ff9c"
-    ],
-    "methods": ["kadena_quicksign_v1"],
-    "events": []
+  "signedCmd": {
```
LindaOrtega, line 393: Is this data structure (the Response type of /sign) defined anywhere? I see it defined at https://kadena-io.github.io/signing-api/#/definitions/SigningResponse, but it doesn't match the properties specified in the KIP

kip-0017.md
jermaine150: Yes that's better. I'll move it to the Request section.
That extra description seems a bit trivial to me, I don't think we need it.

kip-0017.md
jermaine150: The account name is coming from the ledger in a fungible contract, so there's always a fungible in play here.

kip-0017.md
LindaOrtega: Added quick explanation that we're providing Kadena chain ID information in the getAccounts method.

kip-0017.md
LindaOrtega: The Request type in the Haskell code (https://github.com/kadena-io/signing-api/blob/master/kadena-signing-api/src/Kadena/SigningApi.hs#L41) matches the definition in the Open API spec (https://kadena-io.github.io/signing-api/#/definitions/SigningRequest).

The Response type in the Haskell code (https://github.com/kadena-io/signing-api/blob/master/kadena-signing-api/src/Kadena/SigningApi.hs#L60) matches the definition in the Open API spec (https://kadena-io.github.io/signing-api/#/definitions/SigningResponse).

I think some of the discrepancy we're seeing is that pact-lang-api's `signWallet` function (https://github.com/kadena-io/pact-lang-api/blob/fd04d695e6a4ea2203308f25fc781185e0bb3e0a/pact-lang-api.js#L785) formats the `signingCmd` type to match the schema defined above. 

Note the mapping between the `SigningCmd` type in pact-lang-api and what Chainweaver expects (https://github.com/kadena-io/pact-lang-api/blob/fd04d695e6a4ea2203308f25fc781185e0bb3e0a/pact-lang-api.js#L813).

Also, the Response schema has Chainweaver returning the fields `body` and `chainId`. But the `signWallet` function returns just the contents of the `body` field (See https://github.com/kadena-io/pact-lang-api/blob/fd04d695e6a4ea2203308f25fc781185e0bb3e0a/pact-lang-api.js#L788).

PR #34, opened 2023-02-27
KIP-0018: Smart Wallet standard

fmelp:
Putting together a smart wallet standard for wallet builders in our ecosystem and beyond. Note that this is still a draft and format of the KIP may change drastically. First reviews would be specifically for the 'Specification' section.

---
  - https://app.asana.com/0/0/1208668844268866


kraikov:
Does Pact support calling arbitrary contracts/functions? That would be a crucial part for implementing a general purpose smart wallet. Otherwise you would need a smart wallet implementation for the different use-cases - e.g. coin transfer, swap, stake, etc.

edmundnoble:
The first KIP council resolves that this KIP:
- is mostly a product specification for a smart wallet that has not been developed, and not a standard for wallet technology interop. Thus it will be closed.
- there are multiple features in this KIP which have been standardized separately already, #61 and #52.

We expect that wallets and contract developers will make their own developments in this area.

PR #35, opened 2023-03-10
Update the Proposal and Settlement 

Luzzotica:
Without development, the response would actually be invalid, and the WCv2 protocol would reject it (On the wallet side) if it tried to respond with it.

By adding development, it would actually work.


Luzzotica:
Gonna rebuilt more of the proposal and settlement response

PR #37, opened 2023-03-10
KIP-17: SigningCmdRequest -> SigningRequest

Luzzotica:
The Wallet Signing API defines a `SigningRequest`, not a `SigningCmdRequest` object. I feel like having the names match is important here.


LindaOrtega (approved):
Good catch! Thank you for pointing it out.

jermaine150 (approved):

alber70g (approved):

PR #39, opened 2023-03-13
Added types kadena_getAccounts_v1

Luzzotica:
I have given a name to each of the different objects present in the `kadena_getAccounts_v1` just so there is normalcy in communication across languages.

These are the names I have given the objects in my [dart sdk](https://pub.dev/packages/kadena_dart_sdk)

Perhaps adding v1 at the end of each of the models would be a good idea.

### DIFF
--- a/kip-0017.md
+++ b/kip-0017.md
@@ -279,8 +279,8 @@ This method expects a `method` field that is set to `"kadena_getAccounts_v1"`.
 
 This method also expects a `params` object with the following properties:
 
-1. `params`: `Object`\
-    1.1. `accounts`: `Array<Object>`\
+1. `params`: `GetAccountsRequest`\
+    1.1. `accounts`: `Array<AccountRequest>`\
     __ 1.1.1 `account`: `string` - a [WalletConnect Account](#walletconnect-accounts) for which the Kadena account names are requested.\
     __ 1.1.2 `contracts`: `Array<string>` (optional) - array of contracts for which the Kadena account name is requested. Returns all known Kadena accounts when omitted.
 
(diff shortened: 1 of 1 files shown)


LindaOrtega:
Hi @Luzzotica, we chose `Object`, instead of doing named Schemas, for two reasons:
- To keep the syntax of the specifications as close to Javascript as possible
- Be consistent with how other blockchains specify their method RPCs in the WalletConnect docs.
See the following examples:
- https://docs.walletconnect.com/2.0/advanced/rpc-reference/cosmos-rpc#returns
- https://docs.walletconnect.com/2.0/advanced/rpc-reference/solana-rpc#returns

LindaOrtega:
We appreciate the time you put into this, but I'll be closing this PR in 2 days for the above mentioned reasons.

Luzzotica:
> To keep the syntax of the specifications as close to Javascript as possible
> Be consistent with how other blockchains specify their method RPCs in the WalletConnect docs.

Some questions:

1. Why do we want the spec to be as close to Javascript as possible? This feels silly. I'd recommend Typescript at the very least. 
2. Why do we care about being consistent with other blockchains when there is clearly a better route to helping developers building on Kadena?

I recognize how other people are doing it, but I also recognize that there are better approaches that we can take to bring increased clarity to developers working in the ecosystem.

LindaOrtega:
I really appreciate the feedback. I'll answer both of your questions below --

The main thing we're both trying to accomplish is to maximize readability and simplicity of the KIP to make it accessible to as many developers as possible. 

We actually did consider using named schemas instead of `Object`. In keeping with the initial goal of making the definitions accessible to the most developers, we decided to go with `Object` because:
1. It's a concept that's common in multiple programming languages, not just JavaScript.
2. More TypeScript developers might know JavaScript than JavaScript developers know TypeScript.
3. Given that other blockchain projects were also using `Object` in their RPC definitions in the official docs, we also used `Object`. We thought WalletConnect developers that have already implemented other blockchains would appreciate the similarity and make adoption even a tiny bit more likely.
4. Most developers easily associate Object with JSON, which is the final form of the definitions. There might an extra layer of mental association when using a named schema. For example, `<SchemaName> -> what's a <SchemaName>? -> oh is it a JSON structure? -> what's the structure? -> oh its right below` vs. `Object -> oh it's a JSON structure! -> what's the structure? -> oh its right below`.
Screenshot 2023-03-21 at 7 04 01 PM


Given the above reasons, we thought that using `Object` instead of named schema would be be more readable, simple, and accessible to the most developers.

LindaOrtega:
We do want more insight into what developers would actually find readable, accessible, and simple. 

As we continue to work on more complex KIPs and improving the docs website, we want to get as much insight as possible from the community. The PR for these changes and the Tweet announcement are some ways for the community to add feedback. Also, if a developer feels strongly enough about their PR or PR comment, they can do a Twitter poll to see if the rest of the community agrees.

If you have any other ideas on how we can solicit feedback from the community, I'd love to hear it here.

PR #52, opened 2023-11-09
Create KIP-0023: Account management with Passkeys

EnoF:
# Abstract

We propose to use WebAuthn signatures as an alternative option to the current
ED25519 signatures to allow users to approve for transactions processed through
Chainweb Node and Pact smart contracts.

# Motivation

WebAuthn allows for users to use a hardware powered device to store keypairs securely.
Users will be able to interact with those stored keypairs only by initiating a sign
request from the registered web domain. The private key never leaves the device
and the user never enters a password. This brings the user more security and convenience
simultaneously.

In comparison to current wallets, the wallet developers have no access to the
private keys. The user does not have to write down their mnemonics or even enter
their password to decrypt their privatekeys.

**edit 2024-08-06**: Updated the design to use `r:accounts`. This allows us to replace the need of 2 complex smart contracts with just a single straight forward table holding credential pair and leverage the existing security features of an `r:account`. This allows other smart contracts to interact with accounts governed by SpireKey with no additional code as this is no longer governed via a smart contract.

### DIFF
--- a/kip-0023.md
+++ b/kip-0023.md
@@ -10,78 +10,148 @@ Created: 2023-10-27
 
 # Abstract
 
-We propose leveraging WebAuthn signatures as an alternative to ED25519
-signatures for creating passwordless accounts and approving transactions
-processed through Chainweb Node and Pact smart contracts.
+This document describes the proposal to leverage WebAuthn signatures as an
+alternative to ED25519 signatures, paving the way for users to create
+passwordless accounts and approve transactions without managing public and
+private keys. With this proposal, transactions from Pact smart contracts that
+are processed by Chainweb nodes can be signed with biometric authentication.
 
 # Motivation
 
 The current user experience for creating an account and signing transactions can
 be challenging for users unfamiliar with blockchain technology:
 
-- The concept of public / private keypairs is not widely understood.
-- Wallets often generate a 12-24 word mnemonic phrase that users need to store
-  securely in addition to a traditional password.
+- The concept of public and private key pairs is not widely understood.
+- Wallets often generate a 12 to 24 word mnemonic phrase that users need to
+  store securely in addition to a traditional password.
 
-Wallets access their private keys via software. This means that if the wallet is
-compromised, the user's key pairs can be used to sign without the user's
-awareness. With Kadena SpireKey you will always be prompted to personally sign
-for every transaction, which provides more security, transparency and control.
+People often underestimate the importance of keeping their private keys secure,
+and are likely to lose, mistype, or unintentionally expose them. Wallets also
+typically access private keys through software that, if compromised, might be
+used to sign transactions without a user's consent. With Kadena SpireKey and
+WebAuthn signatures, key pairs are stored directly on a secure enclave on the
+user's own devices. Storing keys in this way prevents the private key from being
+exposed or lost. In addition, Kadena SpireKey prompts users to sign for every
+transaction, providing greater security, transparency, and control.
 
-Mnemonic phrases are difficult to store securely and easy to lose. WebAuthn
-enables users to securely generate and store key pairs directly on their own
-hardware devices. There is no need to remember a mnemonic phrase. These key
-pairs can typically be accessed via Touch ID or Face ID, allowing users to log
-in and sign transactions without passwords. Since access is only granted using
-biometric authentication, the need for passwords is removed. Additionally, using
-WebAuthn key pairs provides a user-friendly and convenient experience similar to
-services like Apple Pay or Google Pay.
+Similarly, mnemonic phrases are difficult to store securely and easy to lose.
+With support for WebAuthn signatures, Kadena SpireKey enables users to securely
+generate and store key pairs directly on their own hardware devices, without
+needing to remember or store a mnemonic phrase. The key pairs stored on devices
+can typically be accessed using biometric authentication—like Touch ID or Face
+ID—allowing users to log in and sign transactions without passwords.
+Additionally, WebAuthn has a convienient service provided to sync the key pairs
+to the cloud, in the form of `Passkeys`. This service allows users to make use
+of any `Passkey Provider` of their choice, removing the reliance on big
+corporations to store key pairs.
 
 # Specifications
 
-## Webauthn Pact Keysets
+This section provides technical specifications for the implementation of this
+proposal.
 
-The WebAuthn public key that we provide to Chainweb Node is composed of a JSON
-Web Key (JWK) that has been base64 encoded and prefixed with `WEBAUTHN-`. The
-prefix indicates that different validation needs to be applied in Chainweb Node
-and Pact. This, however, should not impact any smart contracts as WebAuthn
-public keys are accepted as part of any keyset in Pact.
+## Account creation
+
+The `coin` contract removed the ability to rotate keysets since `v6`. This
+change protects users from front-running attacks and squatting attacks. To allow
+new devices to be added retroactively to an account, we need to use a principal
+type that allows keysets to be rotated. This new principal type is called the
+`r:account` principal.
+
+An `r:account` is created from a `keyset-ref-guard`. To protect users from
+front-running and squatting attacks, the `keyset-ref-guard` uses a
+`principal namespaces`. A `principal namespace` is created using a `keyset`, so
+that only the owner of that `keyset` can initialize the namespace on a chain.
+
+A keyset that is used to create the `principal namespace` is not intuitive and
+makes the onboarding more difficult. It's paramount to make sure the user has
+access to the keyset to initialize the namespace on new chains. For this reason,
+one `Passkey` is used as an entropy to create a hierarchical deterministic (HD)
+wallet. A second `Passkey` is created to be included in the keyset used to
+create the `principal namespace` and the `r:account`.
+
+For example:
 
 ```pact
 (env-data
   { 'ks :
     { 'keys :
-    ["WEBAUTHN-a50102032620012158206fb822acf87bea4a37c2d5ff067675456bd38afc4f3d43afd0c7d2c94cd997d6225820c464ff1bccf536172dea9eb37ae3bbfc411bf129afda751ea2f7faace4dbf9c8"]
-    , 'pred : 'keys-all
+      [ "WEBAUTHN-public-key"
+      , "ED25519-public-key-derived-from-HD-wallet-with-Passkey-entropy"
+      ]
+    , 'pred : 'keys-any
     }
   }
 )
-(enforce-keyset (read-keyset 'ks))
 ```
 
-### WebAuthn Keysets
+With the keyset composed of a HD wallet-derived public key and a Passkey, we can
+start the process of creating the `principal namespace` and `r:account`. To
+assist users maintaining their key pairs, we store the credential id's on chain
+in the `kadena.spirekey` contract. This contract allows wallets to keep track of
+the devices that are authorized to sign for transactions and allows the
+management of these devices to be decentralized.
 
-WebAuthn keysets are generated directly on the user's hardware device. This
-process involves creating a public/private key pair where the private key
-remains securely stored on the device, and the public key is shared with the
-service for authentication purposes.
+```pact
+(let* (
+  (ns-name (ns.create-principal-namespace (read-keyset 'ns-keyset)))
+  (ks-ref-name (format "{}.{}" [ns-name 'kadena]))
+)
+  (define-namespace
+    ns-name
+    (read-keyset 'ns-keyset )
+    (read-keyset 'ns-keyset )
+  )
+  (namespace ns-name)
+  (define-keyset ks-ref-name
+    (read-keyset 'ns-keyset)
+  )
+  (let (
+    (account (create-principal (keyset-ref-guard ks-ref-name)))
+  )
+    (coin.create-account
+      account
+      (keyset-ref-guard ks-ref-name)
+    )
+    (kadena.spirekey.add-device-pair
+      account
+      coin
+      { 'guard          :  (read-keyset 'spirekey-keyset) ; keyset of HD wallet derived public key and Passkey
+      , 'credential-id  :  "${credentialId}" ; Credential ID of the device
+      , 'domain         :  "${domain}" ; https://example.com
+      , 'device-type    :  "${deviceType}" ; AAGUID of the device or phone|desktop|security-key if not available
+      , 'color          :  "${color}" ; hex color of the device (optional for user to customize and further identify the device across different wallets)
+      }
+    )
+  )
+)
+```
 
-#### Cloud Backup
+## Transaction signing
 
-When generating keysets without cloud syncing, the private key never leaves the
-user's hardware device. For added convenience, backup and syncing can be enabled
-which involves encrypting the private key on the device before uploading it to a
-cloud service. Because the private key only leaves the device in its encrypted
-form, it remains secure even if the cloud storage is compromised.
+In most cases, users should use the Passkey for the `r:account` to sign
+transactions. This signing method is the most secure way to sign transactions.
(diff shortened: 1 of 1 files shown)


kip-0023.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Pact Keyset
+
+In Pact the public key should be accepted as part of any keyset.
+This allows Smart Contracts to use the registered keysets as any other
+keyset. This should not be impacting any smart contracts, other than
+more types of public keys being accepted.
+
+_NOTE: A keyset using such a base64 encoded public key might exceed_
+_the current limitations of a principaled acount length for newer algorithms_
```
CryptoPascal31, line 113: There is a simple and straightforward solution here.
IMHO: all keysets that include a JWK key (even with single key) should generate `w:` principals. With `w:`, the principal is the hash of the whole keyset.

kip-0023.md
```diff
@@ -0,0 +1,208 @@
...
+  User-->>-WebAuthn(Device): I approve this request
+  WebAuthn(Device)-->>-WalletA.com: Here is the public key: abc000
+  WalletA.com->>+WebAuthn(Device): Here is the transaction to register, please sign
+  WebAuthn(Device)->>+User: Please approve this transaction
+  User-->>-WebAuthn(Device): I approve this transaction
+  WebAuthn(Device)-->>-WalletA.com: Here is the signature for the transaction
+  WalletA.com->>+WebAuthn(SC): Please register "Alice" with this public key
+  WebAuthn(SC)->>+Fungible(SC): Create an account for "c:capabilityguardforalice"
+```
+
```
CryptoPascal31, line 151: I have a very important concern: 
How do you prevent front-running here?
How do you prevent an attacker to impersonate Alice, by registering his own key to the SC just before  Alice ?

raduciobanu22: This is the well known "squatting" issue in that plagued accounts before the introduction of 'k:' accounts

raduciobanu22:
How would the recovery flow look like in case I lose access to the hardware device that I authorized?

kip-0023.md
Randynamic: This is indeed known and something that hasn't been fully hashed out yet. But thank you for the reply 👍🏽

kip-0023.md
raduciobanu22: Wouldn't this impact authorizing a new device?

EnoF: @CryptoPascal31 that is actually a pretty nice idea, the hashing would indeed solve the length problem right away (though I don't think anyone would ever create a `k:` account based on a webauthn key. Reason being that by doing so, you would immediately lock your account to a single domain. 

I believe this would mostly be used for `c:` or `w:` accounts.

kip-0023.md
EnoF: @CryptoPascal31 Yes this indeed is a problem that we already have some idea's around solving. The `alice` alias is used out of convenience reasons, but we already have a PoC running that targets this problem. (Keeping the convenience, but eliminating the squatting problem)

EnoF:
> How would the recovery flow look like in case I lose access to the hardware device that I authorized?

This is a problem that used to exist with "pure webauthn". With the arrival of Passkeys, we are able to use 3rd party key storage to handle those keys. By default users will use their respective OS provided storage, i.e. Google/Apple's solutions, but users can opt in for 3rd party storage like Bitwarden or Lastpass.

Additionally the smart contract can register multiple devices, which would allow for example a backup yubikey to be stored safely in a vault.

kip-0023.md
trendzetter: could you please elaborate how it would be solved?

EnoF: @trendzetter definitely! The "squatting" problem is a bit overloaded with different attack vectors. So I'll address the 3 major concerns we have considered within the team.

### High gas fee steal
This attack vector revolves around stealing an account just before someone wants to register an account by submitting a registration with a different guard with a higher gas fee, resulting in the account ending up belonging to the attacker.
To solve this @enobayram has though of a neat solution. He suggested to implement the registration process using `defpact`, where in the first step you'd provide a hashed value of what you want to register, i.e. your vanity name and guard. In the second step where the account would be minted, you would provide all the data which will be validated against the hash provided in step one. 

### Multi chain registration
Because a vanity account has no link to it's guard, the account could have different guards on different chains. This is normally resolved using `principal named accounts`. However this reduces the convenience of an easy to remember name. 
@javadkh2 has brought up an idea to allow vanity accounts without the above mentioned risk. To solve this problem, we limit on what chain an account can be minted. The minted guards, can only be copied to other chains. To prevent all accounts to be minted on one chain, we suggest to convert the vanity name to an integer and modulo it over the amount of chains we have active currently. The result of that formula is the chain it is allowed to be minted on. I'll leave this up to this point, but we also already have a way to scale up to over 20 chains and how to take care of that. That will likely be a different KIP.

### Pre registration
Then offcourse similar to domain registrations, we will have a group of names that will be very high in demand. Those names could very easily be hogged by a single person. So this could be mitigated in many different ways. You could think of authority based registration, staking, payment, rental, list goes on. 

All the above factors suggestions however are not directly a problem related to WebAuthn. The WebAuthn KIP therefor doesn't really go in depth for this specific problem. Though we do believe that it will provide for a lot of convenience for users and therefor we are already working on several PoC's around those topics. 

Does that take away the concerns with regards to the vanity accounts? cc: @trendzetter @CryptoPascal31 @raduciobanu22

CryptoPascal31: Thank you for the answer. Two remarks:

- For the 2 steps registration, don't forget to impose at the SC level a minimum number of blocks between the 2 steps. Otherwise, it's still vulnerable. A good example is the ENS registration.

- I don't understand why do you want to spread accounts minting on several chains... Why not just define a single minting chain. ?

raduciobanu22: @EnoF Thanks for the explanation! Please check the following scenario, in case I missed something:
I send the step 2 of the defpact, at that point it's visible which account I want to mint. A squatter could then decide he wants to take that account so sends the step 1 tx together with a bunch of spam txs that fill in the block so my step 2 tx isn't mined. Then in the next block, because he sends step 2 with higher gas than I did and he'll mint my account and my tx will fail.

CryptoPascal31: > @EnoF Thanks for the explanation! Please check the following scenario, in case I missed something: I send the step 2 of the defpact, at that point it's visible which account I want to mint. A squatter could then decide he wants to take that account so sends the step 1 tx together with a bunch of spam txs that fill in the block so my step 2 tx isn't mined. Then in the next block, because he sends step 2 with higher gas than I did and he'll mint my account while my step 2 tx will fail.

I think we are on the same wavelength @raduciobanu22. This is exactly the issue I tried to mitigate by enforcing a minimum number of blocks between the 2 steps. It makes this attack a lot harder to conduct and more expensive.

EnoF: > * I don't understand why do you want to spread accounts minting on several chains... Why not just define a single minting chain. ?

Security wise you are absolutely right in that minting on a single chain suffice the use case. The reason for a spread of chain minting is for scalability. With the main idea that any process that can happen on any chain could be "automatically" distributed cross the 20 (or however many chains we have in the future).

With that said, for this specific use case for WebAuthn, it might be even simpler to omit the vanity accounts all together and create a `k:account` based on the first device. Since once your account lives, you could copy over guards from one chain to another, allowing your account to be minted on other chains while your initial device is lost. As long as you have added a second device to your account.

From a user point of view they already have a way to recognize the account by providing a display name. This however loses the ability to share `aliases` with other users, but that is not within the scope of this KIP.

CryptoPascal31: > Security wise you are absolutely right in that minting on a single chain suffice the use case. The reason for a spread of chain minting is for scalability. With the main idea that any process that can happen on any chain could be "automatically" distributed cross the 20 (or however many chains we have in the future).

I'm not sure, it will really help to scale. Because anyway, you will need to submit a transaction to paste the transaction on the chain of interest. And it's not obvious that the "paste transaction" will use much less gas than the initial registration transaction. (I don't even mention that some will refuse to register on chain 13 :smile: )

CryptoPascal31: > With that said, for this specific use case for WebAuthn, it might be even simpler to omit the vanity accounts all together and create a `w:account` based on the first device. Since once your account lives, you could copy over guards from one chain to another, allowing your account to be minted on other chains while your initial device is lost. As long as you have added a second device to your account.

Why ? Your proposal with 2 steps registration procedure should work... At the condition of :
- enforcing a minimum of blocks between two steps (eg 10)
- enforcing on the client side (but maybe on the SC as well) a minimum gas price (eg 1e-3) for the second step.

This would be a great example of leveraging defpacts and commitment/reveal schemes to handle vanity account registration.

FYI : I've just checked. For the ENS registration, they enforce a minimum of 60s (= 5 blocks) between the 2 transactions. At the current gas price, it would cost for an attacker roughly $5000 to front-run and steal an Ethereum name.

EnoF: @CryptoPascal31 the vanity account can still be implemented, but should be done via a separate KIP as WebAuthn does not strictly rely on that concept. 

Also taking into consideration the user experience for first time blockchain users, I'd rather focus on a fast and smooth registration process. So registering the initial device as a `keys-any` resulting in a `w:account` should satisfy both the security and speed of registration. 

The vanity account can be implemented on top of this account, or even as a different smart contract. I agree that it would be very nice to have and we'll definitely be looking into this. (already have PoC's running for this)

raduciobanu22: > @CryptoPascal31 @raduciobanu22 what do you think about using `w:accounts` based on the initial device?
@EnoF w: account is indeed a good option that solves both the security and ux issue. How about adding a 2nd device? That'd imply rotating the keyset and there were discussions about removing this from `w:` accounts

raduciobanu22: > FYI : I've just checked. For the ENS registration, they enforce a minimum of 60s (= 5 blocks) between the 2 transactions. At the current gas price, it would cost for an attacker roughly $5000 to front-run and steal an Ethereum name.

@CryptoPascal31 Did the same math on Kadena? I think the amount is quite affordable :-)

EnoF: @raduciobanu22 Adding a second device on this the webauthn account would not impact the `w:account` name. Simply because we only use the `w:account` to verify initial creation only. That means that if you want to mint your account on a new chain, you could choose to use the initial device (resulting in the same `w:account`) or you could initiate a `defpact` from one of the already existing accounts to copy the guard from one chain to another. 

The guard look up would just look up an array of registered devices(guards) and enforce one of them (or multiple depending on what the contract describes).

daplcor:
Couple of questions/concerns/thoughts on this subject.

I have a slightly different idea/use case for integrating this as on chain multi-factor with a multisig key.  In the sense I'd like to be able to generate smart wallets in multiple ways.  The first way is for twitter users to hit interact with our bot and do #RegisterWallet, I'm handling the wallet generation and mapping it back to twitter oauth.  The second is simply using pact+JS to generate as I am today.  Then with this webauthn, we could register multiple devices(like a business approval workflow use case), so it requires 3 sigs to execute a stored transaction.  Is there any harm in approaching this as a solid MFA that's on chain and we can handle registering devices through a portal linked to the wallet?  

The second thing I'm still trying to wrap my head around is recovery services and how people are potentially exposed in batches from a breach.  That part is still a bit hazy to me and of course, will need to be well covered/documented by each project that supports this.

I'm excited to see where this goes, definitely want the ability to use something like yubikey to store a backup key.  There are secure services that offer products to help vault yubikeys, might be worth trying to bring one of these into the ecosystem?

EnoF:
> Couple of questions/concerns/thoughts on this subject.
> 
> I have a slightly different idea/use case for integrating this as on chain multi-factor with a multisig key. In the sense I'd like to be able to generate smart wallets in multiple ways. The first way is for twitter users to hit interact with our bot and do #RegisterWallet, I'm handling the wallet generation and mapping it back to twitter oauth. The second is simply using pact+JS to generate as I am today. Then with this webauthn, we could register multiple devices(like a business approval workflow use case), so it requires 3 sigs to execute a stored transaction. Is there any harm in approaching this as a solid MFA that's on chain and we can handle registering devices through a portal linked to the wallet?

The WebAuthn generated public keys can be used in keysets, i.e. you could create a keyset consisting of:

```
{ "keys" : 
  [ "webauthn-key-a"
  , "webauthn-key-b"
  , "webauthn-key-n"
  ...
  ]
, "pred" : "keys-2" // or any other pred like a self implemented pred that requires 33% of keys to sign or 3 or whatever your requirements
```

I think from your backend you could perform a local call with `preflight=true` to validate if certain `capabilities` can be brought into scope (i.e. 3 sigs have signed) to assert "consent" and if you want to logged, send the tx and use the tx hash as token? Something along those lines. I have not played with this idea too much, so take it with a grain of salt.

> The second thing I'm still trying to wrap my head around is recovery services and how people are potentially exposed in batches from a breach. That part is still a bit hazy to me and of course, will need to be well covered/documented by each project that supports this.
> 
> I'm excited to see where this goes, definitely want the ability to use something like yubikey to store a backup key. There are secure services that offer products to help vault yubikeys, might be worth trying to bring one of these into the ecosystem?

So yubikeys should be supported by this standard right out of the box. Their backup system is then working as per their specification. With regards to how Google, iCould or BitWarden stores the keys in the cloud is not something I've researched (yet). I believe for most users those services are good enough, specially if used in combination of services/multi devices. 

If I understood Yubikey's solution with Hashicorp's correctly, it allows for a similar setup to the Google/iCould/Bitwarden but then stored on a private server. If that understanding is correct, it implies that the storage of such vaults can be extended by other parties in the future without impact on the implementation proposed.

kip-0023.md
```diff
@@ -0,0 +1,438 @@
...
+be challenging for users unfamiliar with blockchain technology:
+
+- The concept of public / private keypairs is not widely understood.
+- Wallets often generate a 12-24 word mnemonic phrase that users need to store
+  securely in addition to a traditional password.
+
+Moreover, the security of accounts generated through many existing wallets is
+centralized because the user's key pairs are generated and stored by the wallet
+provider. If the wallet provider becomes unavailable or is compromised, users
+may lose access to their assets.
```
CryptoPascal31, line 29: Are you serious, sir ?

EnoF: No this paragraph is inaccurate and is being rewritten.

Randynamic (dismissed):

Randynamic:
@emilypi @jmcardon Could we get your review on this KIP?

kip-0023.md
```diff
@@ -0,0 +1,425 @@
...
+corporations to store key pairs.
+
+# Specifications
+
+This section provides technical specifications for the implementation of this
+proposal.
+
+## Account creation
+
+The `coin` contract removed the ability to rotate keysets since `v6`. This
```
emilypi, line 55:
```suggestion
The `coin` contract removed the ability to rotate the governance predicate (e.g. a keyset) associated with a principal account since `v6`. This
```

emilypi (approved):

Randynamic (approved):

sstraatemans (approved):

trendzetter:
"The coin contract removed the ability to rotate the governance predicate (e.g. a keyset) associated with a principal account since v6. This change protects users from front-running attacks and squatting attacks. "
Afaik this is factually incorrect. Frontrunning doesn't have to do with rotate.

trendzetter:
"The concept of public and private key pairs is not widely understood."
Passkeys are based off public/private key authentication with specific gui and source of entropy is my current understanding so these statements are probably irrelevant.
I like to see your proposal better explained to start with because its hard to have a clear understanding based on the current info while at least everyone with a notion of computer science and even many without CS credentials do understand fundamental concepts such as public/private keys.

PR #53, opened 2023-11-13
KIP 0024 - Proposal of  a new native `(pure)`

CryptoPascal31:
This KIP adds a new native to Pact, that could improve security.

While this is not a silver bullet that will magically get rid of all security issues, the implementation is very straightforward.
In my opinion, it's worth of it.

Initially, this had been proposed here: 
https://github.com/kadena-io/pact/issues/1140

But with more Pact experience, I feel like the (pure) native is necessary.

---
  - https://app.asana.com/0/0/1208668844268878


jmcardon:
This has been implemented in https://github.com/kadena-io/pact-5/pull/367

PR #54, opened 2023-11-17
KIP 0025 : Kadena FS: URLs  

CryptoPascal31:
This KIP propose a new mechanism to reference short data object stored on the blockchain via URLs, 
and associated mechanisms to retrieve them:

This is an early draft and opinions / feedback / ideas are **strongly desired.**

---
As a proof of concept, I deployed several simple storage modules on testnet, and stored many objects:
https://balance.chainweb.com/modules.html?server=api.testnet.chainweb.com&module=free.kdafs-store-one&chain=1

**Simplest:**
kdafs://testnet04:1/free.kdafs-store-one/aGgztKIhm6y8M9iNjHd2txyvyg2ZrVGumP02RtZ-o2A
kdafs://testnet04:14/free.kdafs-store-two/m-l_64Xw86FzD73Nzs9PmxTSXugntZ6dZzTQkt1dgq8

**More complex:**
kdafs://testnet04:1/free.kdafs-store-one/WR1htqVpDlExP1pAcWxysGvODgu2bf5fFPaiWJjX4Sg

**With embedded images:**
kdafs://testnet04:1/free.kdafs-store-one/xZJsVE9hI3OBk1TNF5iSZaMKjxlVhKQ0-pJVSAeRof8
kdafs://testnet04:14/free.kdafs-store-two/LJGE9w87vDl--ShVnMX8mCVgpOWwWqh-TG-98tyYziU

---
I created a simple client dApp that does directly Pact calls, allowing everyone to retrieve above object's URLs.
https://kdafs-dapp.marmalade-ng.xyz/

---
I created and deployed a PoC gateway, that can be used according to the KIP:
https://gw.marmalade-ng.xyz/

For example:
https://gw.marmalade-ng.xyz/kdafs/testnet04:14/free.kdafs-store-two/m-l_64Xw86FzD73Nzs9PmxTSXugntZ6dZzTQkt1dgq8
https://gw.marmalade-ng.xyz/kdafs/testnet04:1/free.kdafs-store-one/WR1htqVpDlExP1pAcWxysGvODgu2bf5fFPaiWJjX4Sg

### DIFF
--- a/kip-0025.md
+++ b/kip-0025.md
@@ -111,7 +111,7 @@ The gateway translates the request to a Chainweb Kadena transaction, submits it 
 The content of ``kdafs://mainnet01:2/nice-namespace.storage-module/9OZ7g0kCU7BRgvgjMD0QLDGUhFOWSuNk7cT6OtTYAw8``
 
 can be retrieved by
-``https://gw.kadena-gw.io/mainnet01:2/nice-namespace.storage-module/9OZ7g0kCU7BRgvgjMD0QLDGUhFOWSuNk7cT6OtTYAw8``
+``https://gw.kadena-gw.io/kdafs/mainnet01:2/nice-namespace.storage-module/9OZ7g0kCU7BRgvgjMD0QLDGUhFOWSuNk7cT6OtTYAw8``
 
 
 **Pact values**:
--- a/kip-0025.md
+++ b/kip-0025.md
@@ -1,0 +1,139 @@
+---
+KIP: "0025"
+Title: Kadena FS and associated URLs
+Author: CryptoPascal
+Status: Draft
+Type: Standard
+Category: General / Pact
+Created: "2023-11-17"
+---
+
+## Abstract
+
+This KIP proposes to specify a Kadena storage standard for small JSON objects, and define ways to retrieve them.
+
+It targets mostly the NFTs metadata, but can be used for many other purposes as well.
+
+## Motivation
+
+Marmalade V1 was storing the token's metadata on-chain. The main advantage was to guarantee that the NFTs metadata had exactly the same lifetime as the NFT ledger's data (ownership, ...).
+
+Marmalade V2 and later NG decided to switch to off-chain storage. This improves the flexibility and completely disconnects the metadata from the ledger's data. But as a drawback, the NFT metadata have a different lifetime dependent on the reliability and long term storage capability of the chosen scheme (ipfs, centralized http server, ...)
+
+The new Ordinals standard on BTC stores all metadata directly on-chain, with a good success story. Such that storing on-chain metadata starts to become the standard.
+
+
+The idea behind this KIP is to unify the best of both approaches:
+- keep flexibility of a Marmalade ledger by only using a "protocol agnostic" URI.
+- propose a URL scheme and access mechanisms, allowing dApp and wallets to retrieve NFTs data similarly as they do for ipfs or https data.
+- give a reliable long term storage possibility for the NFTs creators.
+
+
+## Specification
+
+### `kip.kdafs` interface
+
+
+The following pact interface is proposed: [1]
+
+```pact
+(interface kdafs-v1
+
+  (defun kdafs-immutable:bool ())
+
+  (defun kdafs-get:object (cid:string) )
+)
+```
+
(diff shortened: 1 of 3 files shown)


daplcor:
So, theoretically, this could be applied to more of a general storage environment with lookup capabilities both on and off-chain.  I have advocated for simplification and processes for URIs to move across chains/blockchains.  But, storing the URI in this manner and pointing to an image off-chain is a great way for us to cement the NFT/data in a simple deployment process.  So if a project wanted metadata on chain, they could store it anywhere, say chain 1, for instance, and create either an immutable or non-immutable record.  

This could help enhance games that way, an easy way to upgrade attributes within the single NFT...  The hash process would need to change, and ofc that's up to individual projects, but it's an excellent blueprint for achieving this without complex table storage/mapping.  

Consider the gaming use case as I feel many projects would benefit from a more straightforward mechanism. I assume from reading, we can simply set immutable to false and achieve something similar?  I am curious what would happen with our hash in that case, does the CID remain the same or need to change?

CryptoPascal31:
> This could help enhance games that way, an easy way to upgrade attributes within the single NFT... The hash process would need to change, and ofc that's up to individual projects, but it's an excellent blueprint for achieving this without complex table storage/mapping.
> 
> Consider the gaming use case as I feel many projects would benefit from a more straightforward mechanism. I assume from reading, we can simply set immutable to false and achieve something similar? I am curious what would happen with our hash in that case, does the CID remain the same or need to change?

Thank you for the feedback

The KIP itself **doesn't require** that the data is addressed by its hash. 
It's up to the contract implementation to determine how the CIDs are generated.

The KIP only requires content to be addressed by a unique ID, and the storage module must provide only two functions:
   - ``kdafs-get`` for retrieval
   - ``kdafs-immutable`` to help the gateways/dApps to determine whether the content can be cached. 

---

For my own tests, I created another storage module.. where data is mutable. And the CIDs are randomly generated instead of hashing data. (as you notice, it wouldn't make sense !) 

https://balance.chainweb.com/modules.html?server=api.testnet.chainweb.com&module=free.kdafs-store-mut-one&chain=14

Example of a stored object:
kdafs://testnet04:14/free.kdafs-store-mut-one/d9wkUX2d_SsR-mzB2By27f4P

trendzetter:
I like this idea. Just some thoughts that I had, maybe they aren't applicable or desirable. Would it be useful to have?
- a function to return the max size of an object to store as there is obviously a physical limit dictated by block-size and maybe other factors which may change over time
- a function or attribute to return the file-type and encoding of an object
- some standard for encoding/decoding files. eg base64 -but then it still needs more specification as to which variant-

CryptoPascal31:
>     * a function to return the max size of an object to store as there is obviously a physical limit dictated by block-size and maybe other factors which may change over time

>     * some standardization of the store -write access- function in another interface -to keep it optional- since you compare it to a filesystem

Thank you for your feedback, George.

IMHO, I prefer to restrict the standard to the minimum:
   - A standard way to retrieve objects on chain from the outside.
   - A standard protocol between a module and a gateway from on-chain to outside.
  
Nothing more.

The name may be misleading. I called it FS to highlight some similarities with ipfs:// (especially on the HTTP gateway part). But definitively, this is not a file system.

I prefer letting most freedom possible to the module owner to determine how the objects are stored (in database, in const, ....) with / without restrictions, static, generated .... The maximum size depends on the implementation as well, and how the objects are stored. IMHO this is out of the scope of this KIP proposal.

CryptoPascal31:
>     * a function or attribute to return the file-type and encoding of an object
> 
>     * some standard for encoding/decoding files. eg base64 -but then it still needs more specification as to which variant-
> 

For several reasons, I propose to only standardize and restrict to "objects", and nothing more.
  - It discourages people for doing easily abusing mis-use (like storing a full website on Kadena), by allowing them to store file like it was a new Google Drive.
  - It allows the data to be used on chain too by another module, since they are standards Pact objects, not encoded files.
  - It matches well with the main targeted use case, IMO, NFTs metadata which are always JSON.

As such, types are always "Pact objects" in the Pact world, and JSON in the HTTP world (gateway).

trendzetter:
Just focusing on retrieving basic objects and leaving encoding or other attribute requirements to the application and its needs makes sense in a constrained environment like a blockchain.

PR #55, opened 2023-11-22
KIP-0026: Key Derivation and Mnemonic encoding methods

Takadenoshi:
Introduces a standard for key derivation and encoding mnemonic phrases.

Summary from abstract:

- Derivation: SLIP-10 with derivation path `m/44'/626'/a'`.
- Mnemonics: BIP-39, 24 words, English wordlist, empty passphrase.

Pending: test cases, backwards compatibility section

---
  - https://app.asana.com/0/0/1208668844268886

### DIFF
--- a/kip-0026.md
+++ b/kip-0026.md
@@ -90,7 +90,77 @@ We further suggest using the SLIP-0010 derivation algorithm for the reasons outl
 
 # Test cases
 
-TODO
+## Test 1
+
+BIP39 Mnemonic (English, 24 words)
+
+```
+apple coach sad dial dentist material garden twelve brown plastic business mixed hunt river match pigeon brown tent network enrich dial rigid image milk
+```
+
+BIP39 Seed
+
+```
+bfae0d6049a431cfd8487a8d8a30c5d1c0bbce27f3647aa46ff8bc730c17eaaf109ca36facef9fc06e1bdc26681504c5222fa2c311ca61572cbd244fa856cb59
+```
+
+Derivation paths & public keys
+
+```
+m/44'/626'/0' 2017fee3fb15cfe840e5ed34bf101cc7d5579ffdd20dea09e32fd77c1757f946
+m/44'/626'/1' 58aa1853a0c1e53997c076ff9f274423f9f7548a1c22ea3210af3afeadc557e3
+m/44'/626'/2' 7b3f695ee9ea9036379e78b8d6ca2c9fa8f1dc2c72049b4cdb09967d8ed48c6c
+m/44'/626'/3' a3429e678ffdd3da937519c3c1db785e21917b395a63e0bb9867b2a3701ca81b
+m/44'/626'/4' 949fbf8fdf7ea7883e2e5ae41c3dae44c377b6bdc824251f2348427c5f5aa947
+```
+
+## Test 2
+
+BIP39 Mnemonic (English, 24 words)
+
+```
+first employ other hazard hollow awful helmet fan catch tower own tide used certain access uphold muscle scatter huge october tattoo penalty sugar deer
+```
+
+BIP39 Seed
+
+```
+bed1ef7c826cf185bec95867dded4c9874620c927aa3c7a4ef64fc5e46ba1814fdb9cf2fe012bbb04da38651163c9f74e860eb71a920c852fcb7680f8def4631
+```
+
+Derivation paths & public keys
+
+```
+m/44'/626'/0' 22b65e6ba67a14ca337d4da4bb35515b889943999a870e9215b8472a4dd374b5
+m/44'/626'/1' d88dbea6cb59c1e5442228775fba85ee53269a01860d2ba69ca28ad53517b964
+m/44'/626'/2' 0cef8c1927851176643bde7dffe21e59e491be9be39cfde1e4aadf0fb6b9c0ae
+m/44'/626'/3' 95ea7cc8f5c69de59ff615127c0e055bdb7419b16a3935479f18b198c025719f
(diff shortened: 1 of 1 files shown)


kip-0026.md
```diff
@@ -0,0 +1,105 @@
...
+
+# Motivation
+
+Convergence on these standards will ensure key and mnemonic interoperability between wallet providers. This will benefit users of all wallet types (hot/hardware/cold), who will be able to reuse their keys or phrases without requiring conversion.
+
+# Specification
+
+## Key Derivation
+
+We recommend following the widely adopted standard for deriving ED25519 keypairs: [SLIP-0010](https://github.com/satoshilabs/slips/blob/master/slip-0010.md).
```
jwiegley, line 26: I don't think the KIP itself should be stated as a recommendation. Rather, I'd state directly what anyone following the KIP should follow, and then later we can say to a wallet provider that "We recommend you follow KIP-26".

kip-0026.md
```diff
@@ -0,0 +1,105 @@
...
+24 word mnemonics: support 256 bits of entropy.
+
+The English BIP39 wordlist is recommended for interoperability.
+
+Empty passphrase:
+
+- Allows interoperability.
+  - Few wallets implement encrypted mnemonic recovery.
+- Encrypting a mnemonic with a passphrase may provide a sense of security, but:
+  - A leaked encrypted mnemonic phrase can be quickly brute-forced offline ([BIP-39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#from-mnemonic-to-seed): 2048 round PBKDF2 using salted HMAC-SHA512).
```
CryptoPascal31, line 72: :exclamation: :question:  :open_mouth:
Are you sure that a (good) password derived with a SHA256-2048 PBKDF2 can be **quickly** brute-forced ?

trendzetter: more specifically the assertion that it can be quickly brute-forced is not backed by the reference as far as I can see.

Takadenoshi: Perhaps "significant speed" is more accurate than "quickly", but I think the essence of the statement holds.

A _good_ password could withstand MD5, but it would probably end up looking like another seed phrase to be stored securely, away from the encrypted seedphrase. An average user password would succumb reasonably quickly, especially if the seed was compromised on their machine, where a lot of information could be used to build up a dictionary.

For the current BIP39 spec of HMAC-SHA-512 [1] I estimate that an off-the-shelf 4090 can do about ~1521 kH/s[2] which would map to:

- rockyou.txt wordlist (14M): 9 seconds
- crackstation wordlist (1.2B): 13 minutes
- crackstation * 100 masks (121B): 22 hours
- bruteforce alpha/lower 8 letters (208B): 38 hours
- bruteforce alpha/lower + numeric 8 letters (2.8T): 21 days

And so on. It is not exactly a done-deal given enough entropy, but I'd still call it a fast enough attack with consumer hardware, which will only become cheaper and more accessible in the future[3][4].

I would consider "slow" here to be something like LUKS2/argon2i which is molasses-slow to dictionary attack or bruteforce, or pbkdf2/sha512 with a very large number of rounds (600K+.)

All that said, it should probably read "bruteforced with significant speed", instead of "quickly", so I will change that.

[1] HMAC-SHA-256 would be even faster: A 4090 should do 2048 rounds @ [~4170 kH/s](https://gist.github.com/Chick3nman/32e662a5bb63bc4f51b847bb422222fd#file-rtx_4090_v6-2-6-benchmark-L1592)

[2] Extrapolating from the speed given for 999 iterations [here](https://gist.github.com/Chick3nman/32e662a5bb63bc4f51b847bb422222fd#file-rtx_4090_v6-2-6-benchmark-L1250) of 3120.9 kH/s - rounds to speed should be ~linear

[3] Consider the lastpass situation where some of their early accounts were stuck at < 10K iterations and were cracked after a disasterous leak. Not sure what their hashing function is though.

[4] We don't even know what the "cracking station" state of the art is at the moment. It would not surprise me to learn that there are GPU farms or even custom ASICs that can smoke these numbers.

dfordivam:
@Takadenoshi here is my resposne to your queries
using `m/44'/626'/a'` is ok from the Ledger wallet point of view. Using SLIP-0100 is not the default in Ledger, but is supported. You may also see my prior response on this topic here https://github.com/obsidiansystems/ledger-app-kadena/issues/33

Takadenoshi:
> Using SLIP-0100 is not the default in Ledger, but is supported.

@dfordivam Thanks! 

Q: The not-default part is regarding Ledger not enforcing hardened child keys/segments, right? Is the derivation algorithm is still identical to SLIP-0010 for hardened keys?

dfordivam:
> Is the derivation algorithm is still identical to SLIP-0010 for hardened keys?

No, the algorithm is different and incompatible with the default algorithm (currently in use in ledger-app-kadena)

Randynamic (approved):

kip-0026.md
edmundnoble:
```suggestion
We follow the widely adopted standard for deriving ED25519 keypairs: [SLIP-0010](https://github.com/satoshilabs/slips/blob/master/slip-0010.md).
```

edmundnoble:
The KIP council resolves to accept this KIP as it consists of standard technology that is adopted by wallets across blockchain ecosystems, and so that future wallets can be interoperable as well. In addition, it has already been adopted by the Kadena's Chainweaver 3.

edmundnoble (approved):

javadkh2 (approved):

PR #56, opened 2023-12-10
KIP-0027: Make User Guards SysOnly to ReadOnly

CryptoPascal31:
Following my discussion with @jmcardon,

The proposal is to change the execution environment of User guards from ``SysOnly`` to ``ReadOnly``.

This would be an amazing feature, that would definitively **"unleash the power of Kadena"**

Would like to have the feeling of the team about this. Maybe there are some pitfalls I didn't catch ?

https://github.com/kadena-io/KIPs/blob/06223ea13461aac38702abeefb1e872b2be9cc68/kip-0027.md

---
  - https://app.asana.com/0/0/1208668844268890

### DIFF
--- a/kip-0027.md
+++ b/kip-0027.md
@@ -1,0 +1,155 @@
+---
+KIP: "0027"
+Title: ReadOnly User guards
+Author: CryptoPascal
+Status: Draft
+Type: Standard
+Category: Pact
+Created: 2023/12/09
+---
+
+## Abstract
+
+The Proposal is to allow User guards to read User tables.
+
+
+## Motivation
+
+### Reminder: Purity in Pact.
+
+In Pact, it exists 3 levels of purity environment:
+  - **Impure** (normal)
+  - **ReadOnly**: Writing is not allowed, but Reading is allowed
+  - **SysOnly**: Writing / Reading is not allowed for User tables. Only Sys tables can be read.
+
+Currently (Pact 4.10), User guards called through `(enforce-guard)` must be **SysOnly** pure.
+
+### Rationales and new use cases
+
+Allowing User Guards to be **ReadOnly** pure may allow new use cases, making Pact much more powerful than it actually is.
+
+
+#### Namespace management by DAO / Complex Multi-sigs schemes
+
+Namespaces are managed by 2 guards (User & Admin).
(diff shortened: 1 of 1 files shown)


jmcardon:
So in looking @ this KIP, the only thing I wonder about is: Aren't all of these cases doable via a capability guard that enforces these invariants? That is, take this example:

```
(module balance-checker GOV
  (defcap GOV ()
    (enforce fail "No-Upgrade"))
    
  (defcap HAS_MINIMUM_BALANCE (target-account:string min-balance:decimal) 
      (let ((bal (coin.get-balance target-account)))
      (enforce (>= bal min-balance) "No"))
  ) 

  (defun has-minimum-balance:bool (target-account:string min-balance:decimal)
    (let ((bal (coin.get-balance target-account)))
      (enforce (>= bal min-balance) "No"))
  )

  (defun get-guard:guard (target-account:string min-balance:decimal)
    (create-capability-guard (HAS_MINIMUM_BALANCE target-account min-balance))
  )
  
  (defun my-create-account (target-account:string min-balance:decimal)
    (with-capability (HAS_MINIMUM_BALANCE target-account min-balance)
        (coin.create-account "alice" (balance-checker.get-guard "alice-savings" 1000.0))
  )
)

(coin.create-account "alice" (balance-checker.get-guard "alice-savings" 1000.0))
```

CryptoPascal31:
> So in looking @ this KIP, the only thing I wonder about is: Aren't all of these cases doable via a capability guard that enforces these invariants? That is, take this example:

@jmcardon Your example doesn't work. Or at least is incomplete ...  Where do you acquire the cap ? to unlock the guard. :wink:

More generally, using a cap guard, you always **REQUIRE** to have a caller module to acquire the cap...

Sometimes, it's not very convenient: eg: Alice want's to access her funds by doing a simple ``(coin.transfer )`` at the top level.


But often it's not possible: eg
   - namespace rotation
   - module deployment
   - pact continuation

The first two are necessary to implement namespaces / modules managed by DAO or stuffs like evolved mutli-sigs.
The third is a typical case for a Marmalade buyer.

jmcardon:
This was implemented and merged already

PR #57, opened 2024-02-12
KIP 0028: Pact Verifier Plugins

edmundnoble:
This is a proposal to add Verifier Plugins to Pact as a way to add new cryptography to Pact and integrate it with the capability paradigm. Implementation is at https://github.com/kadena-io/chainweb-node/pull/1777 and https://github.com/kadena-io/pact/pull/1324.

---
  - https://app.asana.com/0/0/1208668844268894

### DIFF
--- a/kip-0028.md
+++ b/kip-0028.md
@@ -28,7 +28,7 @@ We propose defining a *verifier plugin* as a new form of authority in the Pact c
 
 A transaction will be allowed include a list of verifiers, similarly to a list of signers, which for each used verifier will include the name of that verifier, the proof being passed to that verifier, and the capabilities it's being asked to grant. **Note that this means that the creator of a transaction must already know exactly what capabilities a proof grants, to include it into the capability list of a verifier.** This is essential for showing that the things a transaction is allowed to do are included verbatim in the transaction in a human-readable format.
 
-In Pact versions above 4, a capability can require that it was signed for by a given keyset via `enforce-keyset`; analogously, a capability can require that it was granted by a particular verifier via the proposed built-in function `enforce-verifier`, for example `(enforce-verifier 'ZK)`.
+In Pact versions above 4, a capability can require that it was signed for by a given keyset via `enforce-keyset`; analogously, a capability can require that it was granted by a particular verifier via the proposed built-in function `enforce-verifier`, for example `(enforce-verifier 'ZK)` for a verifier named `ZK`.
 
 Some verifiers follow, defined abstractly, which demonstrate how this verifier mechanism is general enough to subsume some existing cryptographic functionality.
 
--- a/kip-0028.md
+++ b/kip-0028.md
@@ -112,8 +112,7 @@ located elsewhere in the transaction, but they are assumed to correspond with
 signers. The first signature corresponds to the first signer, second signature
 to the second signer, and so on.
 
-We propose the following schema for including extra verification types in
-transactions:
+We propose the following schema for a new key in transaction objects:
 
 `verifiers`, an array of objects with:
 
--- a/kip-0028.md
+++ b/kip-0028.md
@@ -1,0 +1,189 @@
+KIP: "0028"
+Title: Pact Verifier Plugins
+Author: Edmund Noble @edmundnoble
+Status: Draft
+Type: Standard
+Category: Interface
+Created: 2023-12-13
+
+# Motivation
+
+Pact allows signers to grant specific capabilities, thereby limiting the *scope* of their signing authorization to only those capabilities that have been granted. For example, a signer can sign a transaction for the capability `(coin.TRANSFER "src" "dest" 1.0)` and their signature can only be taken to authorize a transfer from the "src" account to the "dest" account in the amount of 1.0 KDA; there are no other actions which the signature authorizes. Any attempts within the Pact code of that transaction to grant other capabilities which require signature from the same signer will fail, due to the signature not being *scoped* to those capabilities.
+
+The Pact container, for example chainweb-node, verifies signatures, and informs the Pact environment only that certain keys have signed for certain capabilities. In this way, signatures are invisible to Pact code.
+
+Relatedly, key data is opaque to Pact code; it’s conventionally loaded from per-transaction data, rather than being included literally in Pact source code. A key is opaque data which is associated to capabilities via the Pact container's signature checking.
+
+Put another way, Pact code knows nothing about public key cryptography per se. Pact code instead deals with *capabilities* in the abstract, which can be *attested to* by the Pact container layer as being granted by some *authority*, here the public key of the signer or more generally some keyset. The Pact code in the body of a capability only ensures that it's been granted by the necessary authority.
+
+Other cryptographic algorithms, for example different forms of signature verification, have still occasionally been implemented with Pact code - though there are performance and correctness issues usually associated with implementing them in such a high level language. There has been a steady demand for novel cryptography in Pact driving these efforts. Recently, Pact implemented WebAuthn signature verification for transactions (in addition to the existing ED25519 signature scheme) which fit neatly into Pact's signer/capability paradigm. But verifying signatures on data other than Pact transactions, which may originate from outside Pact - for example from bridges - doesn't easily fit Pact's signer/capability paradigm. Zero-knowledge proofs are another area where new cryptography is useful, which are roughly analogous to signatures for data other than Pact transactions.
+
+Because new cryptography is needed, an interface for developers to easily integrate their own cryptographic algorithms with the capability-oriented paradigm of Pact would be very valuable.
+
+Such an interface is proposed here.
+
+# Design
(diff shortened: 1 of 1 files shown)


jwiegley:
Excellent initial exposition, @edmundnoble, you've made the intent of the proposal very clear.

CryptoPascal31:
**Suggestion :** The REPL simulation function ```(env-verfiers  )``` should be described (or at least mentioned) in the KIP.

**Question 1:** Is a managed cap automatically installed by a verfier (like it is for a signature) ?

**Question 2:** Does the magic triggered by GAS_PAYER cap works with verifier ? I mean, can we trigger a gas station from a verifier ?

**Question 3:** Do you plan to create a new type a guard that automatically enforce a verifier ? (similar to a keyset guard)

**Question 4:** Do you plan to add some doc for the verifiers (at least one short paragraph ) ? btw: it may answer to my question 1.

edmundnoble:
@CryptoPascal31 thanks for the questions!

> Suggestion : The REPL simulation function (env-verfiers) should be described (or at least mentioned) in the KIP.

Good idea, I'll add that.

> Question 1: Is a managed cap automatically installed by a verfier (like it is with a signature) ?

Yes.

> Question 2: Does the magic triggered by GAS_PAYER cap works with verifier ? I mean, can we trigger a gas station from a verifier ?

Not initially. This is mentioned in the KIP:

"A verifier may charge gas after gas is bought and before any transaction code is run. Because of this, currently, a verifier cannot grant the `GAS_PAYER` capability, which would allow paying for gas using a verifier’s attestation. This may change in the future."

> Question 3: Do you plan to create a new type a guard that automatically enforce a verifier ? (similar to a keyset guard)

No plans to do this at the moment. I would recommend using a capability guard for this.

> Question 4: Do you plan to add some doc for the verifiers (at least one short paragraph ) ? btw: it may answer to my question 1.

What exactly are you asking about documenting? The Pact code, chainweb code, this KIP, or something else? Or is there a particular workflow you're asking about documenting?

CryptoPascal31:
Thank you for your answers @edmundnoble 

> > Question 2: Does the magic triggered by GAS_PAYER cap works with verifier ? I mean, can we trigger a gas station from a verifier ?
> 
> Not initially. This is mentioned in the KIP:
> 
> "A verifier may charge gas after gas is bought and before any transaction code is run. Because of this, currently, a verifier cannot grant the `GAS_PAYER` capability, which would allow paying for gas using a verifier’s attestation. This may change in the future."

In fact, I was especially thinking about this: 
https://github.com/kadena-io/chainweb-node/issues/1702

I was wondering whether the "Allow" verifier could be an elegant solution for the suggestion I raised. Currently, a GAS_PAYER triggered gas station is not related to the signature, as it may not be related not to a verifier attestation. I know, I'm not clear..
But maybe you should take a quick look at the code (function `findPayer` of `TransactionExec.hs` . The code just searches for GAS_PAYER cap in the signatures and if found it calls the related module, to unlock the guard of the sender ...  independently of who signed the cap... I mean, maybe it would be possible to search into the "verifiers caps" in addition to the "signatures caps". 



> > Question 4: Do you plan to add some doc for the verifiers (at least one short paragraph ) ? btw: it may answer to my question 1.
> 
> What exactly are you asking about documenting? The Pact code, chainweb code, this KIP, or something else? Or is there a particular workflow you're asking about documenting?

I mean in the official Pact doc: 
https://github.com/kadena-io/pact/blob/master/docs/en/pact-reference.md

There is a lot about caps, signatures, managed caps... I think the concept of verifiers should be explained here.

IMHO: new users are not supposed to know that the documentation is spread between the official doc and the KIPs.. I mean, users are not supposed to search into the KIP repository to understand the concept of a verifier.

kip-0028.md
```diff
@@ -0,0 +1,174 @@
...
+
+We propose the following schema for including extra verification types in
+transactions:
+
+`verifiers`, an array of objects with:
+
+- `name`:
+    - an identifier for the verification type.
+        - a string.
+- `proof`:
```
javadkh2, line 127: Is the proof a string or JSON? Since the example is a string, but here it says JSON

kip-0028.md
```diff
@@ -0,0 +1,174 @@
...
+
+Because new cryptography is needed, an interface for developers to easily integrate their own cryptographic algorithms with the capability-oriented paradigm of Pact would be very valuable.
+
+Such an interface is proposed here.
+
+# Design
+
+We propose defining a *verifier plugin* as a new form of authority in the Pact container layer that can grant capabilities. More specifically, we propose defining it as a named function which, given a proof value and a set of capabilities, either consents to grant that set of capabilities exactly or produces an error.
+
+A transaction will be allowed include a list of verifiers, similarly to a list of signers, which for each used verifier will include the name of that verifier, the proof being passed to that verifier, and the capabilities it's being asked to grant. **Note that this means that the creator of a transaction must already know exactly what capabilities a proof grants, to include it into the capability list of a verifier.** This is essential for showing that the things a transaction is allowed to do are included verbatim in the transaction in a human-readable format.
```
LindaOrtega, line 29: Some (nit) wording suggestions.
```suggestion
A transaction will be allowed to include a list of verifiers, similarly to a list of signers. Each verifier will include the name of that verifier, the proof being passed to that verifier, and the capabilities it's being asked to grant. 

**Note that this means that the creator of a transaction must already know exactly what capabilities a proof grants, to include it into the capability list of a verifier.** 

This is essential for showing that the things a transaction is allowed to do are included verbatim in the transaction in a human-readable format.
```

kip-0028.md
```diff
@@ -0,0 +1,174 @@
...
+
+In Pact versions above 4, a capability can require that it was signed for by a given keyset via `enforce-keyset`; analogously, a capability can require that it was granted by a particular verifier via the proposed built-in function `enforce-verifier`, for example `(enforce-verifier 'ZK)`.
```
LindaOrtega, line 31: More nit suggestions:
- Some more wording suggestions
- And if it would be possible to add an example of what `ZK` in `(enforce-verifier 'ZK)` is referring to.
 
```suggestion
In versions above Pact 4, a capability can require that it was signed for by a given keyset via `enforce-keyset`; analogously, a capability can require that it was granted by a particular verifier via the proposed built-in function `enforce-verifier`. For example `(enforce-verifier 'ZK)`, where `'ZK` is `<todo>`.
```

kip-0028.md
```diff
@@ -0,0 +1,174 @@
...
+
+## `chainweb-node`
+
+In `chainweb-node` we propose adding a mapping from verifier names to verifier functions, which is consulted when validating a transaction. A verifier may charge gas after gas is bought and before any transaction code is run. Because of this, currently, a verifier cannot grant the `GAS_PAYER` capability, which would allow paying for gas using a verifier’s attestation. This may change in the future.
+
+Notably, writing a verifier plugin does not require changes to `pact`, only to `chainweb-node`, because verifiers run on the Pact container layer and verifier names are strings.
+
+## Changes to the Pact transaction schema
+
+For reference, this is the schema of the `signers` field:
```
LindaOrtega, line 84: Whats the connection between `signers` and `verifiers` field? Is it included here to show that both have similar type structures?

kip-0028.md
larskuhtz: I think, that the proof string just be just binary. In the JSON body of the tx it would be encoded as base64url (without padding).

Proof strings messages from the outside world that don't have a meaning in Pact. We should constrain the content as little as possible. So plain bytes seem the right choice. 

It is the job of the verifier to understand those plain bytes and relate them to the Pact semantics of the content of the `clist`.

kip-0028.md
```diff
@@ -0,0 +1,189 @@
...
+transactions:
+
+`verifiers`, an array of objects with:
+
+- `name`:
+    - an identifier for the verification type.
+        - a string.
+- `proof`:
+    - input to the verifier.
+        - an arbitrary JSON-encoded Pact value.
```
larskuhtz, line 125:
```suggestion
        - JSON string with the base64Url (without padding) encoded proof binary data.
```

Randynamic (approved):

Randynamic:
@edmundnoble I believe we can resolve the conversations and merge this proposal.

kip-0028.md
edmundnoble: The proof string right now is indeed a JSON representation of a Pact value. We can change this later in a subsequent KIP.

kip-0028.md
edmundnoble: Yes that's all.

kip-0028.md
edmundnoble: We can absolutely change this in a subsequent patch or KIP.

jmcardon (approved):

jmcardon (approved):

LindaOrtega (approved):

PR #58, opened 2024-02-14
KIP-0029: Proposal of new guard type: Constant boolean guard

CryptoPascal31:
The idea is to propose a new guard, the most simple as possible, the easiest to understand.

But it's currently missing, and would be very useful, and would improve the clarity.
Nothing revolutionary.

https://github.com/CryptoPascal31/KIPs/blob/119fe2df7df6a42bba7f46c406ca631aec95e99a/kip-0029.md

### DIFF
--- a/kip-0029.md
+++ b/kip-0029.md
@@ -1,0 +1,92 @@
+---
+KIP: "0029"
+Title: Constant Boolean Guards
+Author: CryptoPascal
+Status: Draft
+Type: Standard
+Category: Pact
+Created: 2024-02-14
+---
+
+## Abstract
+
+The Proposal is to add a new type of guards in Pact:
+**Constant boolean guards**.
+
+
+## Motivation
+
+Very often it's necessary to create "always success" or "always fail" guards.
+Some examples are:
+- Definition of free but non-rotatable namespaces.
+- Default guards in smart-contracts.
+- Burn accounts
+- Temporary accounts
+- and more ...
+
+Currently the main adopted workaround is to use an user guard with a function that:
+ - returns true for "Always success"
+ - enforce false for "Always fail"
+
+But the workaround is not perfect. A user has to dig into the module code, verify whether the module is upgradable. And it definitively lacks of clarity for most people.
+
+
+We need an unambiguously way to express, identify and give a guarantee of immutability of such conditions.  
+
+## Specification
+
+#### Pact function
+The proposed interface for the new native is as follows:
+
+```pact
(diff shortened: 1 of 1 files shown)


daplcor:
This seems extremely useful, I assume it would help with things like the function below.  You can type specify bool, arrays, objects, strings, but not guards that I am aware of.  So we end up having to add extra helper functions which serve no real purpose.     
(defcap USERPAY (account:string)
    (with-default-read taxes account {'guard: (create-null-guard) } {'guard:=guard}
      (if (!= guard (create-null-guard)) 
        (enforce-guard (at 'guard (coin.details account)))
        (enforce-guard guard)
      )
    )
   )
   
   (defun enforce-null:bool
    ()
  false)

  (defun create-null-guard:guard
    ()
  (create-user-guard (enforce-null)))

jmcardon:
Apologies for the late reply on this one, but this one is achievable with userland code.

```pact
(module m g (defcap g () true)
  (defun enforce-bool (b:bool) (enforce b "boolean guard failed"))
  (defun create-boolean-guard:guard (b: bool) (create-user-guard (enforce-bool b)))
  ```

CryptoPascal31:
> Apologies for the late reply on this one, but this one is achievable with userland code.
> 
> ```pact
> (module m g (defcap g () true)
>   (defun enforce-bool (b:bool) (enforce b "boolean guard failed"))
>   (defun create-boolean-guard:guard (b:bool) (create-user-guard (enforce-bool b)))
> ```
Yes of course it is.
But IMHO, such a common and simple need should be handled and guaranteed by the language itself. 

Plus, having an associated clean principal (as I propose) would improve a lot of transparency and security for non-technical users. Everybody should be able to identify quickly Burn / "Free to use" accounts.

Because with a userland workaround, anybody who wants to verify what's going on:
  - need to check the associated userland function, 
  - check the code of the function.
  - check the governance and upgrade conditions of the module...

I can give you an example:
Everybody thinks that the `free` namespace is "free"... But everybody is wrong, In fact:
   - It's guard given by `describe-namespace` is a user guard to `ns.success`
   - While the function `ns.success` is correct, the module `ns` can be upgraded by `ns-admin-keyset` (probably Kadena :wink:)  
   - Basically, it means that the owners of these keys could revoke the authorization of the `free` userspace at anytime. => I believe this is obviously not considered... :smiley:

Having a constant boolean guard, managed by Pact, would allow fulfilling such use-cases more easily and without any possible ambiguity.

PR #59, opened 2024-02-16
KIP-0030: SpireKey dApp <> Wallet Connection

EnoF:
Scoped KIP down to dapp <> wallet connection

---
  - https://app.asana.com/0/0/1208668844268904

### DIFF
--- a/kip-0030.md
+++ b/kip-0030.md
@@ -1,6 +1,6 @@
 ---
 KIP: "0030"
-Title: Spirekey Connect
+Title: Kadena Spirekey Connect
 Author: Andy Tang @EnoF
 Status: Draft
 Type: Standard
--- a/kip-0030.md
+++ b/kip-0030.md
@@ -1,0 +1,207 @@
+---
+KIP: "0030"
+Title: SpireConnect
+Author: Andy Tang @EnoF
+Status: Draft
+Type: Standard
+Category: Interface
+Created: 2024-01-22
+---
+
+# Abstract
+
+We propose to use a different way for dApps to connect with Wallets using
+existing and proven Web2 Standards in order to improve the usability for end
+users interacting with dApps and Wallets.
+
+# Motivation
+
+As of writing there are several ways to connect a dApp to a Wallet in order to
+sign for transactions. We believe that the available setups do not provide for
+the seamless connection we ambition for.
+
+On the Web2 space there are well established techniques to provide authorization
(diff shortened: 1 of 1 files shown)


Randynamic (dismissed):

Randynamic (approved):

kip-0030.md
```diff
@@ -0,0 +1,227 @@
...
+>;
+export type Account = {
+  alias: string;
+  accountName: string;
+  minApprovals: number;
+  minRegistrationApprovals: number;
+  balance: string;
+  devices: Device[];
+  guard?: Guard;
+  keyset?: Keyset;
```
emilypi, line 193: Is this guard significantly different from the keyset? It would be good to explain the difference.

EnoF: I'll update the supporting text around it. :+1:

alber70g:
As per discussed with @Randynamic and @EnoF I propose to combine this with KIP-0017 and create a new KIP as an extension. I even think we can use this more widely. 

I would propose to make a new KIP describing **Spirekey Connect**. 

**Spirekey Connect** is a comprehensive messaging interface that can then be used on various transport layers, like WalletConnect, URL-#-based (eg `wallet.kadena.io/spirekey?returnURL=mydapp.com/spirekey#<message>`), endpoint based `POST <host>/spirekey`, `window.spirekey` or `window.kadena.spirekey(message)`.

That way we can separate the transport layer from the messaging layer and generalize the way dapps communicate with Kadena wallets.

The messages give support for:
* `connect`: open the wallet and return keys (and accounts?)
* `getAccounts`: retrieve accounts relevant to a certain key
* `signWithRequirements`: a message that contains a finished transaction (similar to Chainweaver v2s /quicksign) and the requirements that the wallet needs to fulfill to complete that transaction (like moving fungibles across chains, etc)
* `transactionRequest`: a message that contains a piece of code with an incomplete transaction (similar to the current /sign of chainweaver v2) . The code would have `read-x` statements that, based on some additional metadata, can be filled by the wallet by presenting a form based on the metadata.

Note: `connect` and `getAccounts` are not a hard requirement. A dApp can also request the account from the user from within the dapp, and submit a `signWithRequirements` or `transactionRequest` afterwards.

alber70g:
Next to this, we should understand what we mean with SpireKey. What is it? There's currently too much noise:  
Is it a wallet, signing using WebAuthn, using a smart contract to make smart wallets governed by a WebAuthn key, is it r-accounts?

PR #61, opened 2024-08-14
KIP-0031 - Batch Signing Transactions w/ WebAuthn

imalsogreg:
Adds a KIP for batch-signing transactions with the WebAuthn protocol.

[Rendered](https://github.com/kadena-io/KIPs/blob/kip-0031/kip-0031/kip-0031.md)


kip-0031/kip-0031.md
```diff
@@ -0,0 +1,157 @@
...
+    "signature": BASE64_URL,
+    "clientDataJSON":  BASE64_URL,
+    "authenticatorData": BASE64
+  },
+  "merkleProof": BASE64_URL
+}
+```
+
+These “batch signatures” can be included with transactions as ordinary
+signatures, in the "sigs" field. The transactions in a batch will each have the
```
emilypi, line 94: This should probably link to the yaml format for API requests. 

```suggestion
signatures, in the "sigs" field of a Pact transaction API request. The transactions in a batch will each have the
```

kip-0031/kip-0031.md
```diff
@@ -0,0 +1,157 @@
...
+In the example above, the process is:
+
+- Leaves A and B are connected, creating the first inner node 1.
+- Leaves C and D are connected, creating node 2.
+- Node 1 is connected to node 2, creating node 3.
+- Nodes E and F are connected, creating node 4.
+- There are no remaining leaves at this point, so the remaining nodes are connected to form a tree.
+- Nodes 3 and 4 are connected, creating node 5. Node 5 is the root of the resulting tree.
+
+Each leaf node with a transaction hash `h` has the value `blake2b_256(0x00 <> h)`.
```
emilypi, line 127: you might want to explain the `<>` syntax

emilypi:
So i have a few hesitations about this, and i think what's missing is how it's going to be integrated specifically into the chainweb nodes. Like, we need examples of the following (in my view):

- How it's going to work with /send and the other endpoints if it's going to be a yaml format change?
- What are the precise semantics of batching, which were a little obfuscated in the text: like, if we send 20 transactions to the chain via batch, how independent are they going to need to be? How does mempool select the right ones? Are they executed in a block in order, or are we taking the merkle root as gospel?
- It seems to imply that the transactions will be executed one after another, but that means a mempool change currently, and that needs to be fleshed out

Conceptually, I like the idea as a feature, but it violates some ownership principles pretty onerously: a transaction batch delegates its final execution authority to a mysterious process in the nodes, presumably in the mempool, which makes this a difficult story to support. I'm also left with the question of: why this can't exist as a contract? For those reasons, I think we're going to have to go back to the drawing board and really flesh this story out and motivate it properly. 

I'm closing for now, due to the above.

edmundnoble:
For the record, the PR for this is https://github.com/kadena-io/pact-5/pull/179.

PR #63, opened 2024-09-28
KIP-0031: Fungible token Metadata

CryptoPascal31:
The goal of this KIP is 
  - to propose a standard way for token issuers to include meta-data in their token modules codes.
  - a way to allow wallets, frontends ...  to retrieve and use these data

This KIP is an early draft. I'm not sure that all needs are covered.
**Reviews, pros, cons, comments, improvements, and suggestions are welcome and desired**  

GH Formatted Markdown document link: 
https://github.com/CryptoPascal31/KIPs/blob/kip-031/kip-0031.md

GH formatted interface code:
https://github.com/CryptoPascal31/KIPs/blob/kip-031/kip-0031/fungible-meta-v1.pact

And example:
https://github.com/CryptoPascal31/KIPs/blob/kip-031/kip-0031/example-token.pact

EDIT: Added 3 new paragraph to address questions

---
  - https://app.asana.com/0/0/1208668844268908

### DIFF
--- a/kip-0031.md
+++ b/kip-0031.md
@@ -1,0 +1,91 @@
+---
+KIP: "0031"
+Title: fungible-meta-v1
+Author: CryptoPascal  
+Status: Early Draft
+Type: Standard
+Category: Chainweb / Fungible
+Created: 2024-09-28
+---
+
+## Abstract
+
+Define and implement a new way for tokens creators to export their Meta-Data to Wallets, Frontends, and Other services.
+While this is currently handled off-chain, this proposal is intended to take a an "on-chain" approach.
+
+
+## Rationale
+
+While `fungible-v2` provides technical functions to allow "technical on-chain" interoperability between tokens, meta-data and information about fungible tokens are missing.
+
+Currently, listing tokens for Wallet, Frontends, Exchanges, and other third parties is a cumbersome process. Maintainers have to manually contact tokens issuers to retrieve logos, names, and other useful information to display to users. This ends up with disparities between different wallets or frontends. Users complain that they don't have reliable information about their investment.
+
+Moreover, token issuers have no clean and defined way to inform wallets when some changes happen with their token (Rebrand, new issuance, ...)
+
+And finally, since Kadena is multichains, wallets need to be informed on how to use the token, regarding chains deployment:
+  - Some tokens may not be available on all chains.
+  - Some tokens may have a single issuance chain where supply is accounted
+
+The proposal is **to create an optional interface that can be implemented by token modules** to automatically inform third parties about their meta-data, and displayable information to users. Wallets, frontends, ... are expected to make local calls to the `kip.fungible-meta` implemented functions.
+
+
+
+### API
+
+The interface `kip.fungible-meta`, which should be implemented by token modules, provides the following functions.
+
+##### (general-info)
+ -> *object:{kip.fungible-meta.info-schema}*
+
+Return general information about a token:
+  - `symbol` *string*: Token Ticker symbol. Usually a short name in uppercase.
+  - `name` *string*: Token name. Usually a relative short name composed of 1 or 2 words.
+  - `description` *string*: Longer description of the token. Not mandatory
+  - `img` *string* : URL of the graphical representation of the token. Not mandatory
+  - `logo` *string* : URL of a small (res max of 192x192) graphical representation of the token. Not mandatory. Can be the same as `img`
+  - `color` *string*: A CSSv4 compatible color (https://www.w3.org/TR/css-color-4/#typedef-color). Not mandatory. Can be used by wallets or frontends for some graphical displays (charts ...)
+  - `socials` *[object:{kip.fungible-social-schema}]*: A list of tuples (`type` *string*, `url` *string*). The list of allowed types (non limitative): "discord", "twitter", "website", "telegram", "reddit", "github"  
+
+**Note:** All optional fields must use an empty string if the information is not provided.
+
+##### (main-chain)
+  -> *string*
+
+Return the main chain (if any) of the token. Usually, it's the issuance chain. This chain is expected to have the most reliable information about total and circulating supply. All compliant wallets should request this chain before displaying supply information to the user.
+
+##### (supported-chains)
+  -> *[string]*
+
+Return the list of chains supported by the token. This is important for wallets, for retrieving balances, and correctly managing available X-chain transfers.
+
+
+##### (total-supply)
+  -> *decimal*
+
+Return the total Supply of the token. If unknown 0.0 can be returned.
+
+##### (circulating-supply)
+  -> *decimal*
+
+Return the total circulating supply. If unknown, 0.0 can be returned. This information might be relevant only on the main chain.
+
+**Important note:** All those functions implementations must be "Read-Only pure". They are not allowed to write data or have any side effects.
+
+
+## Backwards Compatibility
+
+Since this interface may be implemented in addition to `fungible-v2`, "non compliant" tokens will continue to work as usual.
+Hence, no backward compatibility issues are expected.
+
+
+## Specification
+
+See [fungible-meta-v1.pact](kip-0031/fungible-meta-v1.pact)
+
+## Example
+
+Implementation example:
+  See [example-token.pact](kip-0031/example-token.pact)
+
+## References
+* KIP-0005: Fungible v2
--- a/kip-0031.md
+++ b/kip-0031.md
@@ -28,6 +28,22 @@ And finally, since Kadena is multichains, wallets need to be informed on how to 
 
 The proposal is **to create an optional interface that can be implemented by token modules** to automatically inform third parties about their meta-data, and displayable information to users. Wallets, frontends, ... are expected to make local calls to the `kip.fungible-meta` implemented functions.
 
+#### Trustability
+
+Information (especially supply) relies **only** on token issuers and module writers good faith. Clients should not blindly trust them. Sorting legitimate tokens from illegitimate ones is not in the scope of this KIP.
+
+#### Non-Obligability
+For being compliant with this KIP, token issuers have no obligation to provide all data.
+And clients (Wallets, Frontends), have no obligation to show everything.
+
+#### Supply considerations
+
+One important piece of information required by investors is the total and circulating supply.
+Kadena being multichain, when dynamic minting/burning tokens are considered, a new issue arises:
+  - Usually, minting / burning / emission happens on a single chain.
+  - **But other chains are not aware**: they just receive X-chain transfers.
+
+As a consequence, it's highly probable that supply information won't be accurate on most chains. That's why this KIP proposes the concept of **main-chain** => The module itself informs the client where an accurate value of supply amounts can be found.
 
 
 ### API
--- a/kip-0031/example-token.pact
+++ b/kip-0031/example-token.pact
@@ -1,0 +1,47 @@
+(load "fungible-meta-v1.pact")
+
+(module bro-token G
+  (defcap G ()
+    (enforce false "No"))
+
+  (implements kip.fungible-v2)
+  (implements kip.fungible-meta-v1)
+
+
+  (defconst _SUPPORTED_CHAINS ["0", "1", "2" ])
+  (defconst _MAIN_CHAIN "2")
+
+  (defun general-info:object{info-schema} ()
+    {'symbol: "BRO",
+     'name: "$BRO",
+     'description: "Token of Kadena Brothers DAO",
+     'img: "https://raw.githubusercontent.com/brothers-DAO/bro-token/refs/heads/main/graphic-assets/basic/BRO_3000_3000.png",
+     'logo: "https://raw.githubusercontent.com/brothers-DAO/bro-token/refs/heads/main/graphic-assets/basic/BRO_192_192.png",
+     'color: "#eec2a1",
+     'socials:[{'type:"website", 'url:"https://bro.pink"},
+               {'type:"github",  'url:"https://github.com/brothers-DAO"}]}
+  )
+
+  (defun main-chain:string ()
+    _MAIN_CHAIN)
+
+  (defun supported-chains:[string] ()
+    _SUPPORTED_CHAINS)
+
+  (defun total-supply:decimal ()
+    100.0)
+
+  (defun circulating-supply:decimal ()
+    (if (= _MAIN_CHAIN (at 'chain-id (chain-data)))
+        (- (total-supply) (get-balance "issuance-account"))
+        0.0)
+  )
+
+  ; Usual fungible-v2 functions ===>
+
+  ; ...
+
+  ; ...
+
+  ; ....
+)
--- a/kip-0031/fungible-meta-v1.pact
+++ b/kip-0031/fungible-meta-v1.pact
@@ -1,0 +1,34 @@
+(interface fungible-meta-v1
+  @doc "Provides several meta-data informations for the token"
+
+  (defschema social-schema
+    type:string
+    url:string
+  )
+
+  (defschema info-schema
+    symbol:string
+    name:string
+    description:string
+    img:string
+    logo:string
+    color:string
+    socials:[object{social-schema}]
+  )
+
+  (defun general-info:object{info-schema} ()
+    @doc "Return general information about a token")
+
+  (defun main-chain:string ()
+    @doc "Return the main chain (if any) of the token")
+
+  (defun supported-chains:[string] ()
+    @doc "Return the list of chains supported by the token")
+
+  (defun total-supply:decimal ()
+    @doc "Return the total supply of the token")
+
+  (defun circulating-supply:decimal ()
+    @doc "Returns the circulating supply of the token")
+
+)


daplcor:
I am not sure 100% about circulating supply, it might be easier/better to have a trusted ecosystem project run a daily or weekly update service for this but max supply is definitely good to capture.  Worth a wider discussion at the very minimum.  I'd love to hear from wallets, dexs, and other services that pull in token information.

CryptoPascal31:
> I am not sure 100% about circulating supply, it might be easier/better to have a trusted ecosystem project run a daily or weekly update service for this but max supply is definitely good to capture. Worth a wider discussion at the very minimum. I'd love to hear from wallets, dexs, and other services that pull in token information.

The idea behind this is to have "Best effort approach". Nothing is mandatory, and based of good faith and interpretation of tokens issuers. It let the responsibility for tokens to report the information themselves. Several cases could happen:
   - Some tokens may have always total-supply = circulating-supply => They just return the same value
   - Some tokens may decide that circulating-supply is not relevant, or impossible to estimate => They return 0.0
   - Some tokens contracts may have appropriate counters and/or special internals to account the supply => They can use `(circulating-supply)` to return the exact value to users. (eg $BRO token, or future Linx wrapped tokens)
   
=> But in all cases, it's at least better than nothing.

Note that this is ,not incompatible with third party scripts, that could provide this information too, and spot scammers. Anyway 3rd party projects may have some limitations for exacting data if they are not aware of the internals of the token. Think for example about $BRO, whose the "non-circulating" supply is kept on a special account... and a dumb script won't be aware of that.

daplcor:
I believe that's fine as we are talking best case for circulating supply and something is better than nothing.  I would however like to consider removing the color field, as I don't feel like it adds anything.

LindaOrtega:
Hi, in general this KIP has received positive internal feedback from the Kadena team. I'll re-post some of the feedback one of the team members gave that summarizes the main points:

> After reading the KIP, my understanding is that the current metadata fields include:
> 
> - symbol: Token Ticker symbol
> - name: Token name
> - description: Longer description (optional)
> - img: URL of graphical representation (optional)
> - logo: URL of small graphical representation (optional)
> - color: CSSv4 compatible color (optional)
> - socials: List of social media links (optional)
> 
> I would propose adding a mandatory 'decimals' field to specify the number of decimal places a token supports.
> 
> - Critical for wallets and interfaces to properly display token amounts
> - Essential for calculating and handling token precision accurately
> - Consistent with other token standards (like ERC20) where decimals is standard metadata
> - Prevents display errors and would reduce user confusion
> 
> Considerations:
> 
> - How chainweb upgrades would affect metadata implementation
> - Field standardization needs, including:
> - Maximum lengths for text fields (symbol, name, description)
> - Format requirements for social media URLs
> 
> Other relevant constraints
> 
> - Migration path for existing tokens created before this standard, including:
> - How to handle legacy tokens
> - Process for upgrading to the new standard

Some additional feedback and notes:
1. This interface needs to be versioned in case we want to upgrade it in the future.
2. The migration strategy should provide examples of how token issuers can upgrade their contract to implement this interface. And how they would migrate if we put out future versions of it.
3. The migration strategy should provide guidance to governance-bound tokens (aka tokens whose smart contract cannot be upgraded) as well.
4. The current fungible-v2 interface has a `precision` constant, should we duplicate this field in this interface as well?
5. We should also discuss if we want to upgrade the coin contract to implement this interface. Though it's not a blocker for this KIP.

Logistical planning (open for discussion):
1. If we think this interface is a good idea, we should deploy it to the `kip` interface.
2. We should have a Testnet period for this KIP where users can try out this interface before we launch to Mainnet.

alber70g:
I have some reservations on this kip

1. if this is a KIP (which I think is too intrusive given that it's a Kadena improvement proposal, not a community standard (on separate note, should we have something that represent community standards? Like KIP-0017, or KIP-0026 draft)), should we make this for coin as well? If we do, how will `(circulating-supply)` be calculated? Is it as simple as `height * reward`? Do we have this information at pact-runtime level?
2. Is there something relatable outside Kadena ecosystem that we can refer to?
3. If not 2, why should this be something we want to do as an optional thing. Wouldn't it be better to enforce this and have a new fungible-v3 include this metadata

Practical considerations
4. both `img` and `logo`?
5. There's socials, but not website? Is that also a social?
6. How do we prevent users from creating fake symbols? Do we want to guard it with the guard that guards the fungible?
7. what determines the "main-chain" for a token to provide this information from?
8. Should `(supported-chains)` be a boolean, `isChainSupported` on the chain where the query is executed?

CryptoPascal31:
> Hi, in general this KIP has received positive internal feedback from the Kadena team. I'll re-post some of the feedback one of the team members gave that summarizes the main points:
> 
> > After reading the KIP, my understanding is that the current metadata fields include:
> > 
> > * symbol: Token Ticker symbol
> > * name: Token name
> > * description: Longer description (optional)
> > * img: URL of graphical representation (optional)
> > * logo: URL of small graphical representation (optional)
> > * color: CSSv4 compatible color (optional)
> > * socials: List of social media links (optional)
> > 
> > I would propose adding a mandatory 'decimals' field to specify the number of decimal places a token supports.
> > 
> > * Critical for wallets and interfaces to properly display token amounts
> > * Essential for calculating and handling token precision accurately
> > * Consistent with other token standards (like ERC20) where decimals is standard metadata
> > * Prevents display errors and would reduce user confusion
> > 
The fungible-v2 already has a `(decimals)` function. IMO. It shouldn't be duplicated.

**Note:** Decimals in Kadena have a different meaning and usage  than in ERC-20. 
   - The EVM doesn'r have the decimal type, and all balances are stored in retrieved in Integers. So the DECIMALS variables is critical to display and handle amounts properly, from a user point of view. But for SC, decimals are transparent.
   -  In Kadena, (decimals) is mainly used by SC to round properly amounts, before calling transfer functions.


> > Considerations:
> > 
> > * How chainweb upgrades would affect metadata implementation

Smoothly, as soon chainweb preserves general backward compatibility for Smart Contracts.

> > * Field standardization needs, including:
> > * Maximum lengths for text fields (symbol, name, description)

Yes.

> > * Format requirements for social media URLs

Yes

> > 
> > Other relevant constraints
> > 
> > * Migration path for existing tokens created before this standard, including:
> > * How to handle legacy tokens
> > * Process for upgrading to the new standard
> 
> Some additional feedback and notes:
> 
>     1. This interface needs to be versioned in case we want to upgrade it in the future.
> 
Yes it is... That's why the proposed interface is called `fungible-meta-v1`

>     2. The migration strategy should provide examples of how token issuers can upgrade their contract to implement this interface. And how they would migrate if we put out future versions of it.
Yes, I can add this in the KIP Markdown document, but I'm not sure that the right place for such guidance is in the Standard itself.. Maybe in an annexed document ?

> 
>     3. The migration strategy should provide guidance to governance-bound tokens (aka tokens whose smart contract cannot be upgraded) as well.
> 
In this case there are 2 possibilities: either:
  - Don't upgrade: Keep as is.
  - Create a new "proxy" module that calls old fungible module + add meta-data.

>     4. The current fungible-v2 interface has a `precision` constant, should we duplicate this field in this interface as well?
> 
As I wrote above: IMO no.

>     5. We should also discuss if we want to upgrade the coin contract to implement this interface. Though it's not a blocker for this KIP.
Nothing mandatory. But IMO this is not necessary, as the coin contract should be kept at a minimum... And the coin contract is already well known by all wallets and Dapp..


> 
> 
> Logistical planning (open for discussion):
> 
>     1. If we think this interface is a good idea, we should deploy it to the `kip` interface.
Yes
> 
>     2. We should have a Testnet period for this KIP where users can try out this interface before we launch to Mainnet.
Of course

**Important note:** The philosophy of this KIP is not to enforce a mandatory upgrade... From a “technical” point of-view, `fungible-v2` will be the only mandatory and minimal interface for tokens to work and interact with 3rd party SC. The objective of this KIP is to only propose a better user-experience through wallets and Dapp, by providing informative (ie meta) "data"...

CryptoPascal31:
> I have some reservations on this kip
> 
> \1. if this is a KIP (which I think is too intrusive given that it's a Kadena improvement proposal, not a community standard (on separate note, should we have something that represent community standards? Like KIP-0017, or KIP-0026 draft)), 

I don't know... It depends on the internal policy of Kadena.

> should we make this for coin as well? 

IMO No. The coin contract should be kept at a minimum... And the coin contract is already well known by all wallets and Dapp..

> If we do, how will `(circulating-supply)` be calculated? Is it as simple as `height * reward`? Do we have this information at pact-runtime level? 

> 2. Is there something relatable outside Kadena ecosystem that we can refer to?

ERC-20 provides only 2 optional fields for meta-data: `name` and `symbol`.. But this doesn't mean, that we can't do better in Kadena.
FYI: I was inspired by the meta-data used by the Ecko-DEX frontend.

>  3. If not 2, why should this be something we want to do as an optional thing. Wouldn't it be better to enforce this and have a new fungible-v3 include this metadata

Because IMO, we should separate two things:
  - `fungible-v2` which is the “technical interface”, which is the minimum and **mandatory** interface to be able to interact with 3rd party SC and Wallets.
  - `fungible-meta-v1` (this KIP): Nothing technical, just informative... Improve the Used-Experience and helps to make easier the maintenance of Wallets and Dapp frontends.  


> 
> Practical considerations 4. both `img` and `logo`?

`img` is supposed a general purpose (maybe high resolution image). While `logo` is supposed to be a low-res image that can be fetched quickly.. when (most of the time) only a small image is needed.

>  5. There's socials, but not website? Is that also a social? 
 
Yes... It's included in `socials` (read the KIP carefully) but maybe we should add another speific field for this, or maybe rename `socials` to a more general term (like `external-links`)  ? 
 
>  6. How do we prevent users from creating fake symbols? Do we want to guard it with the guard that guards the fungible?
 
Nothing we can do on a public blockchain. Trademark enforcement is not in the scope of a blockchain. It's up to the user to enter trusted tokens modules refs in his wallet.. Or it's up to the wallet's maintainers to list safe tokens.


 > 7. what determines the "main-chain" for a token to provide this information from?

It's up to the token issuer to decide wether the token has a main-chain, and what is the main-chain. For dynamically minted token (eg token related to a bridge), this will be the chain where tokens are minted: the "chain who knows the supply information" 


 > 8. Should `(supported-chains)` be a boolean, `isChainSupported` on the chain where the query is executed?

The idea behind `(supported-chains)`, is to allow wallets to be quickly aware of the list of supported (btw probably synonym to deployed ) chains, by doing a single query on a single chain (eg chain 0). Imagine if Kadena increases the number of chains to 10.000, querying every chain to see whether the token is deployed won't be possible.

openaccess-id:
**- Data Overhead:** Data only needs to be written to a single chain's contract (e.g. register on first 20 chains)
= Enforces data is on certain chains, this reduces the number of network requests required to locate the main chain (e.g., searching the first 20 chains rather than later 1000 chains).

**- Supported Chain Ranges:** Chains should support configurable ranges, e.g., 0-4 or 9-19.
(Supported chains are only useful for abstraction of simple contracts/dapps) (headsup)

hswopeams:
>  2. Is there something relatable outside Kadena ecosystem that we can refer to? 
We could refer to the Ethereum ([ERC-721](https://eips.ethereum.org/EIPS/eip-721)) metadata standard. Search in the browser on "metadata".

It is also an optional standard. It uses a JSON schema that has most of the metadata properties in a `properties` object that doesn't have specific property names. This makes it flexible so that properties can be added without having to modify the schema.

kip-0031.md
```diff
@@ -0,0 +1,107 @@
...
+Author: CryptoPascal  
+Status: Early Draft
+Type: Standard
+Category: Chainweb / Fungible
+Created: 2024-09-28
+---
+
+## Abstract
+
+Define and implement a new way for tokens creators to export their Meta-Data to Wallets, Frontends, and Other services.
```
emilypi, line 13:
```suggestion
Define and implement a new way for tokens creators to export their Metadata to Wallets, Frontends, and Other services.
```

kip-0031.md
```diff
@@ -0,0 +1,107 @@
...
+While this is currently handled off-chain, this proposal is intended to take a an "on-chain" approach.
+
+
+## Rationale
+
+While `fungible-v2` provides technical functions to allow "technical on-chain" interoperability between tokens, meta-data and information about fungible tokens are missing.
```
emilypi, line 19:
```suggestion
While `fungible-v2` provides technical functions to allow "technical on-chain" interoperability between tokens, metadata and information about fungible tokens are missing.
```

kip-0031.md
```diff
@@ -0,0 +1,107 @@
...
+
+Currently, listing tokens for Wallet, Frontends, Exchanges, and other third parties is a cumbersome process. Maintainers have to manually contact tokens issuers to retrieve logos, names, and other useful information to display to users. This ends up with disparities between different wallets or frontends. Users complain that they don't have reliable information about their investment.
+
+Moreover, token issuers have no clean and defined way to inform wallets when some changes happen with their token (Rebrand, new issuance, ...)
+
+And finally, since Kadena is multichains, wallets need to be informed on how to use the token, regarding chains deployment:
```
emilypi, line 25:
```suggestion
And finally, since Kadena is consists of multiple chains, wallets need to be informed on how to use the token, regarding chain deployment:
```

kip-0031.md
```diff
@@ -0,0 +1,107 @@
...
+  - Some tokens may not be available on all chains.
+  - Some tokens may have a single issuance chain where supply is accounted
+
+The proposal is **to create an optional interface that can be implemented by token modules** to automatically inform third parties about their meta-data, and displayable information to users. Wallets, frontends, ... are expected to make local calls to the `kip.fungible-meta` implemented functions.
```
emilypi, line 29:
```suggestion
The proposal is **to create an optional interface that can be implemented by token modules** to automatically inform third parties about token metadata and displayable information to users. Wallets, frontends, ... are expected to make local calls to the `kip.fungible-meta` implemented functions.
```

kip-0031.md
```diff
@@ -0,0 +1,107 @@
...
+And clients (Wallets, Frontends), have no obligation to show everything.
+
+#### Supply considerations
+
+One important piece of information required by investors is the total and circulating supply.
+Kadena being multichain, when dynamic minting/burning tokens are considered, a new issue arises:
+  - Usually, minting / burning / emission happens on a single chain.
+  - **But other chains are not aware**: they just receive X-chain transfers.
+
+As a consequence, it's highly probable that supply information won't be accurate on most chains. That's why this KIP proposes the concept of **main-chain** => The module itself informs the client where an accurate value of supply amounts can be found.
```
emilypi, line 46: Supply will be accurate on individual chains if there as a chain-level supply noted in the interface for a token's metadata, just not a "total supply" across all the network. I think here, we need to separate the idea of what is _autonomous_ metadata, as in facts that a ledger can keep about itself, vs _dependent_ metadata - facts that require some kind of oracle to contribute. A main chain being an example of this oracle, but not the only valid kind out there.

kip-0031.md
```diff
@@ -0,0 +1,107 @@
...
+
+The interface `kip.fungible-meta`, which should be implemented by token modules, provides the following functions.
+
+##### (general-info)
+ -> *object:{kip.fungible-meta.info-schema}*
+
+Return general information about a token:
+  - `symbol` *string*: Token Ticker symbol. Usually a short name in uppercase.
+  - `name` *string*: Token name. Usually a relative short name composed of 1 or 2 words.
+  - `description` *string*: Longer description of the token. Not mandatory
```
emilypi, line 59: proposing a maxchar cap on this to avoid spamability

kip-0031.md
```diff
@@ -0,0 +1,107 @@
...
+  - `img` *string* : URL of the graphical representation of the token. Not mandatory
+  - `logo` *string* : URL of a small (res max of 192x192) graphical representation of the token. Not mandatory. Can be the same as `img`
+  - `color` *string*: A CSSv4 compatible color (https://www.w3.org/TR/css-color-4/#typedef-color). Not mandatory. Can be used by wallets or frontends for some graphical displays (charts ...)
+  - `socials` *[object:{kip.fungible-social-schema}]*: A list of tuples (`type` *string*, `url` *string*). The list of allowed types (non limitative): "discord", "twitter", "website", "telegram", "reddit", "github"  
+
+**Note:** All optional fields must use an empty string if the information is not provided.
```
emilypi, line 65: which ones are optional? Is this the same as "not mandatory"? Why are `socials` mandatory?

emilypi (commented):
Looks thank you @CryptoPascal31. Adding some comments here.

LindaOrtega:
@CryptoPascal31 did you have a chance to look at @emilypi's feedback?

alber70g:
@CryptoPascal31 I have read your reply. All the things I didn't mention I'm good with. The things addressed below I have a comment about

> > Practical considerations 4. both `img` and `logo`?
> 
> `img` is supposed a general purpose (maybe high resolution image). While `logo` is supposed to be a low-res image that can be fetched quickly.. when (most of the time) only a small image is needed.

Let's make this clear by using descriptive names names, like: `img` `img-lg`

> > 5. There's socials, but not website? Is that also a social?
> 
> Yes... It's included in `socials` (read the KIP carefully) but maybe we should add another speific field for this, or maybe rename `socials` to a more general term (like `external-links`) ?

Let's make it general term, like: `urls`?

> > 7. what determines the "main-chain" for a token to provide this information from?
> 
> It's up to the token issuer to decide wether the token has a main-chain, and what is the main-chain. For dynamically minted token (eg token related to a bridge), this will be the chain where tokens are minted: the "chain who knows the supply information"

Agreed! Good to consider non-native tokens, e.g. bridged ones

> > 8. Should `(supported-chains)` be a boolean, `isChainSupported` on the chain where the query is executed?
> 
> The idea behind `(supported-chains)`, is to allow wallets to be quickly aware of the list of supported (btw probably synonym to deployed ) chains, by doing a single query on a single chain (eg chain 0). Imagine if Kadena increases the number of chains to 10.000, querying every chain to see whether the token is deployed won't be possible.

It could also consider all chains where this token resides? Similar to 7.

aserranoni:
Guys? Is this going to be pushed forward in any capacity?

PR #65, opened 2024-11-05
KIP-0032: Cross-Chain Function Call Syntax Enhancement

DaisukeFlowers:
This KIP proposes a syntax enhancement for Pact that enables specifying the chain ID within function calls to facilitate direct cross-chain interactions. By including the chain ID in function calls, we aim to simplify and expand cross-chain workflows, reducing reliance on yield/resume constructs for data transfer only. This will allow contracts to directly interact across chains, improving developer efficiency and usability for complex multi-chain operations.

Looking forward to feedback and any suggestions on refining the proposal!

---
  - https://app.asana.com/0/0/1208690299578288

### DIFF
--- a/kip-0032.md
+++ b/kip-0032.md
@@ -44,7 +44,7 @@ Kadena's current cross-chain communication utilizes `defpact`, `yield`, and `res
 ```pact
 (defpact crossChainTransfer
   (step "Step 1" (yield { "data-key": data-value }))
-  (step "Step 2" (resume { "data-key": data-value })))
+  (step "Step 2" (resume { "data-key":= data-value })))
 ```
 
 ### Proposed Cross-Chain Function Call
--- a/kip-0032.md
+++ b/kip-0032.md
@@ -54,6 +54,15 @@ Kadena's current cross-chain communication utilizes `defpact`, `yield`, and `res
 
 This allows developers to call `functionZ` directly on `chain123` within `namespaceX.contractY`.
 
+### Example
+```pact
+;; Current usage
+(free.token.get-balance ACCOUNT)
+
+;; Proposed usage
+(1.free.token.get-balance ACCOUNT)
+```
+
 ## Backward Compatibility
 
 The proposal is fully backward-compatible. The existing `(namespace.contract.function PARAMETERS)` syntax remains unchanged for single-chain calls.
--- a/kip-0032.md
+++ b/kip-0032.md
@@ -1,5 +1,5 @@
 ---
-KIP: "0029"
+KIP: "0032"
 Title: Cross-Chain Function Call Syntax Enhancement by Including Chain ID in Function Calls
 Author: @DaisukeFlowers
 Status: Draft
@@ -8,7 +8,7 @@ Category: Pact
 Created: 2024-11-04
 ---
 
-# KIP-0029: Cross-Chain Function Call Syntax Enhancement by Including Chain ID in Function Calls
+# KIP-0032: Cross-Chain Function Call Syntax Enhancement by Including Chain ID in Function Calls
 
 ## Abstract
 
--- a/kip-0029.md
+++ b/kip-0029.md
@@ -1,0 +1,65 @@
+---
+KIP: "0029"
+Title: Cross-Chain Function Call Syntax Enhancement by Including Chain ID in Function Calls
+Author: @DaisukeFlowers
+Status: Draft
+Type: Improvement
+Category: Pact
+Created: 2024-11-04
+---
+
+# KIP-0029: Cross-Chain Function Call Syntax Enhancement by Including Chain ID in Function Calls
+
+## Abstract
+
+Introducing a syntax that includes the chain ID in the function call to enable direct cross-chain interactions. This proposal aims to simplify cross-chain workflows by allowing function calls across chains, which is currently limited to data transfer only. With this change, cross-chain communication becomes more intuitive and flexible, making the development of complex interactions more efficient.
+
+## Motivation
+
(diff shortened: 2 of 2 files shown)


kip-0032.md
```diff
@@ -0,0 +1,74 @@
...
+
+# KIP-0032: Cross-Chain Function Call Syntax Enhancement by Including Chain ID in Function Calls
+
+## Abstract
+
+Introducing a syntax that includes the chain ID in the function call to enable direct cross-chain interactions. This proposal aims to simplify cross-chain workflows by allowing function calls across chains, which is currently limited to data transfer only. With this change, cross-chain communication becomes more intuitive and flexible, making the development of complex interactions more efficient.
+
+## Motivation
+
+Kadena's current cross-chain communication utilizes `defpact`, `yield`, and `resume` constructs to transfer data between chains. While effective for data exchange, this approach is limited and can become complex for developers who need to execute functions or retrieve information from other chains directly.
```
LindaOrtega, line 19: Hi @DaisukeFlowers, I'm rewording this a bit to make sure I understand the main ask:
- Is this asking for smart contracts to call functions in other chains?
- This would be similar to how smart contracts can currently call functions in other smart contracts within the same chain, correct?

LindaOrtega: _**Example of reading the state in another chain:**_
Let say the following Pact code is being **executed on Chain 0**:
```Pact
(free.token.get-balance ACCOUNT_A)
(1.free.token.get-balance ACCOUNT_A)
```
`(free.token.get-balance ACCOUNT_A)` will retrieve the account balance of `ACCOUNT_A` in Chain 0.
But `(1.free.token.get-balance ACCOUNT_B)` will retrieve the account balance of `ACCOUNT_A` in Chain 1.

_**Example of changing the state in another chain:**_
Let say the following Pact code is being **executed on Chain 0**:
```Pact
(2.free.token.transfer ACCOUNT_A ACCOUNT_B 2.5)
(free.someOtherToken.transfer ACCOUNT_B ACCOUNT_A 1.0)
```
`(2.free.token.transfer ACCOUNT_A ACCOUNT_B 2.5)` will transfer funds from these accounts on chain 2.
`(free.someOtherToken.transfer ACCOUNT_B ACCOUNT_A 1.0)` will transfer funds from these accounts on chain 0.

-----

Are these examples correct?

DaisukeFlowers: Hi Linda, all statements are correct. 
The function will be will executed as in the stated chain.
In the last example: 
If the following is run in the chain 0
`(2.free.token.transfer ACCOUNT_A ACCOUNT_B 2.5)`   
ACCOUNT_A from chain 2 will transfer to ACCOUNT_B from chain 2. 
The function will run as if in chain 2 from chain 0.

LindaOrtega: Aw, I see! 

**TL;DR:**
Unfortunately, being able to execute code on multiple chains (especially code that changes the state) is not feasible at the moment with the current `chainweb-node` architecture. 

That means that chain X cannot run code on chain Y directly.

**Longer Explanation:**
Each chain has its own block and Pact history. You can think of the different chains in Kadena as "mini" blockchains in their own right; and together they make up the bigger Kadena network. Its just that the consensus algorithm of the entire network requires hash information from all chains (see the [Whitepaper](https://cdn.sanity.io/files/agrhq0bu/production/1e4f5f6b99d1cb0701432df6e23e9bbbf46ea363.pdf) for more details).

So for the different chains to communicate with each other about their respective Pact state, they need to use a "bridge" (of sorts). This "bridge" is what we usually call `cross-chain` operations.

------

But I would like to explore the problem you brought up a little bit more, and see if there is another solution that can work. For example, there might be a way to make using `defPact` easier. 

Could you elaborate on the use case you had in mind that needed the feature being proposed? And is it something that's feasible (but complex) to do right now?

DaisukeFlowers: I understand. 
What I'm trying to accomplish is the possibility to pay dividends. 
In order to do that, all shares contracts in all chains need to know how much KDA they have to calculate the dividend per share.
If the shares contract in chain 0 has 50 KDA, shares contract in chain 1 has 50 KDA and shares contract in chain 2 has 0 KDA, and all my shares are in contract in chain 2, even if it has 0 KDA, I would still be entitle of the corresponding of the total 100 KDA that the company holds.
In order to do this, each chain needs to transmit what they hold to all other chains and once all chains know how much they hold they can calculate how much dividend per share is, and do the analysis of how much they have and how much they owe, and then if they owe more than they have, all shares contracts in different chains will need to balance and transfer between them to have enough to pay the dividends they owe. 
To do all this, is a big amount of cross-chain transfers and blocks. 
It would be easier if I could just read the tables in other chains, but that will not be possible.
The other option is to do it in 1 chain, I have that contract 100% functional but I really want to do it multi-chain.
I just need to find a good way to test all this cross-chain interaction ( looking forward cross-chain in RELP, I hope we get it soon)
Thanks you @LindaOrtega

edmundnoble: You can absolutely do this multi-chain. But this will require typical techniques from bridges and more generally distributed systems, like locking up shares for some period so that the chains can all agree for that period.

DaisukeFlowers: Hi @edmundnoble, I found it better to calculate the time weighted average of each share as part of the transfer function and just use that number to calculate the dividend per share than having a locking system. 
Once I can test and make sure everything works as intended in cross-chain I would love to share it with you and get some feedback.

edmundnoble: Sure. Time-weighting sounds feasible, just as long as you've calculated that the time-weight is sufficient to prevent any way you'd game the system by doing other things with the shares in the meantime.

edmundnoble:
The second KIP council resolves to close this KIP as it's infeasible to do this as-written in a way that gives the desired semantics.

We recognize though that there are areas in which there's some desire for improvement in the multi-chain developer experience, whether that's in Pact's syntax, or in making additional types of low-cost communication possible between chains.

PR #66, opened 2024-11-26
kip-0033: Lightweight Confirmation Layer

openaccess-id:
This proposal introduces a lightweight confirmation layer for Kadena, aimed at improving the scalability, reliability, and performance of multi-chain applications.


openaccess-id:
While this concept could be expanded further, the proposed approach is designed to minimize network load while providing the critical tools necessary for highly scalable and redundant projects. By utilizing these features, projects can confidently ensure secure, uninterrupted user experiences and maintain high throughput.

PR #68, opened 2024-11-27
Create kip-0034.md

openaccess-id:
This proposal suggests adding a WebRTC endpoint to Chainweb nodes for text-based communication. The goal is to distribute workload, improve efficiency, and replace the WalletConnect approach. The WebRTC WebSocket would establish peer-to-peer connections between browser-based dApps (local users) and wallets (local) for secure text-only communication. 

The Chainweb node would handle and validate the initial connection setup only, enhancing privacy and enabling large-scale transaction data sharing directly between peers, bypassing relay servers.

### DIFF
--- a/kip-0034.md
+++ b/kip-0034.md
@@ -1,0 +1,145 @@
+---
+KIP: 0034
+Title: "KIP-0034: Include WebRTC WebSocket/Signal Server in Chainweb Node (Prepare for WalletConnect Replacement)"
+Author: openaccess @openaccessX
+Status: Draft
+Type: Standard
+Category: P2P/Networking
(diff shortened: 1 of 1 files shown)


openaccess-id:
Regarding KIP 0030 – WebRTC as a Unified Standard for dApp-Wallet Communication:

- Same device: WebRTC can be used with a local signal handler to avoid the need of cross-domain handling (though window.postMessage is neat).
- Cross-device: WebRTC can also function with a deployed signaling server (as proposed in the nodes).
For the sake of developer UX, why not implement a single modular, extendable, and minimal solution instead of two separate implementations?

PR #69, opened 2024-11-27
KIP-0033: Lightweight Confirmation Layer for Kadena (e.g., DAG or Oth…

openaccess-id:
…er Technologies)

This proposal introduces a lightweight confirmation layer for Kadena, aimed at improving the scalability, reliability, and performance of multi-chain applications. The layer would enhance data efficiency in Chainweb while preserving its core multi-chain architecture and finality.

A confirmation layer would help Kadena handle high demand, support seamless multi-chain dApps, and improve overall system throughput.

### DIFF
--- a/kip-0033.md
+++ b/kip-0033.md
@@ -1,0 +1,93 @@
+---
+KIP: 0033
+Title: Lightweight Confirmation Layer for Kadena (e.g., DAG or Other Technologies)
+Author: openaccess @openaccessX
+Status: Draft
+Type: Standard
+Category: P2P/Networking / Chainweb (Soft) / Layer
+Created: 2024-11-26
+---
+
+# KIP 0033: Lightweight Confirmation Layer for Kadena (e.g., DAG or Other Technologies)
+
+## Summary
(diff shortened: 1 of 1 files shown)


openaccess-id:
Note: This does not need to be part of the initial implementation unless a clear path is established.

### Additional Consideration After Proposal Implementation (Brainstorming):
Following the implementation of this proposal, a potential remaining challenge to throughput reliability could be "gas wars," where transactions prioritize inclusion in blocks by offering higher gas prices after confirmation. While this mechanism increases transaction costs, it also poses risks, such as targeted attacks to diminish the functionality of competitors. For example, large companies might view attack costs as justified if the resulting reputational or operational damage to competitors yields financial gains exceeding the attack cost.
_(Example for large companies/whales: attack cost = competitor impact + reputational damage; potential gains from a full takedown = increased earnings or stock shorting profits exceeding the attack cost.)_

### Possible Options/Considerations/Solutions:
**- Unify Gas Prices:** Eliminate gas-based competition for block inclusion, because the dApps are able to route transactions to less congested chains.
**- Dedicated Block Space Priority:** Reserve a fixed portion of block space (e.g., 50,000 gas) for small transactions, default coin transfers, and cross-chain references to ensure their timely processing without delays.
**- Restrict High-Gas Transactions:** Permit only basic coin transfers and cross-chain transactions to offer higher gas prices for priority. Cap their block space usage at a set percentage (e.g., 50%) unless they are first-in-time rather than prioritized by gas price.

PR #71, opened 2025-03-04
Feat/rpc methods WIP

nillo:
WIP


kip-0037.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Backwards Compatibility
+
+- Legacy Methods: Wallets MAY support older methods for retrieving account information alongside kadena_getAccounts_v1. Legacy methods MUST be documented as deprecated and phased out over time.
+- Configuration Consistency: Implementations MUST ensure that the returned account information matches the wallet’s active configuration.
+
+## Security Considerations
+
+- Data Accuracy: Ensure that the returned account information is accurate and verified.
+- Event Notification: Emit account change events to keep dApps updated and prevent configuration mismatches.
```
alber70g: Modify this and reference the optional KIP that describes events. In that KIP you can add that WalletConnect supports events, and the wallet-adapter will implement those events on a best-effort basis. E.g. when a wallet provider has support for event-based communication this can then propagate to the dapp, or can be emulated by the wallet-adapter itself

kip-0038.md
```diff
@@ -0,0 +1,213 @@
...
+            "keys": [
+              "35d381861d9a6cfb84bef47930b386d1a3b4f3742c8e87772a9e6b43c0e7b5f7"
+            ],
+            "pred": "keys-all"
+          }
+        }
+      ]
+    }
+  ]
+}
```
alber70g, line 135: Should we include a second contract in the example?

kip-0038.md
```diff
@@ -0,0 +1,213 @@
...
+### Response details
+
+Accounts should be returned per fungible contract, if the wallet knows about multiple fungibles held by one account then return the account for each one.
+
+If a wallet only works with single accounts, this method should return a single account.
+
+### Request by public key
+
+kip-0017 defines `kadena_getAccounts_v1`, which uses a public key received from walletconnect to retrieve accounts. The v2 method maintains compatibility with this approach.
+
```
alber70g, line 160:
```suggestion
[KIP-0017](https://github.com/kadena-io/KIPs/blob/master/kip-0017.md) defines `kadena_getAccounts_v1`, which uses a public key received from walletconnect to retrieve accounts. The v2 method maintains compatibility with this approach.

```

kip-0038.md
```diff
@@ -0,0 +1,213 @@
...
+**Method Name:** `kadena_getAccounts_v2`
+
+### Request Parameters
+
+```typescript
+interface GetAccountsRequest {
+  accounts?: Array<{
+    publicKey: string;
+    contract: string;
```
alber70g, line 33:
```suggestion
    publicKey?: string;
    contract?: string;
```
See also comment on KIP-0037

kip-0039.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Response
+
+The method returns a NetworkInfo object containing details about the active network:
+
+```typescript
+interface NetworkInfo {
+  name: string // The human-readable name of the network.
+  networkId: string // The unique identifier for the network (e.g., "mainnet01").
+  url?: string // Optional: The network endpoint URL.
```
alber70g, line 47: Specify what this endpoint represents. Whether it should be just `api.chainweb.com` or the full path to where the commands can be sent eg.: `https://api.chainweb.com/chainweb/0.0/mainnet01/chain/1/pact/api/v1`.
Should this contain placeholders for chainId?

kip-0039.md
```diff
@@ -0,0 +1,134 @@
...
+}
+```
+
+### Response
+
+The method returns a NetworkInfo object containing details about the active network:
+
+```typescript
+interface NetworkInfo {
+  name: string // The human-readable name of the network.
```
alber70g, line 45:
```suggestion
  networkName: string // The human-readable name of the network.
```
Use `networkName` as key for this?

kip-0040.md
```diff
@@ -0,0 +1,129 @@
...
+### Response
+
+The method returns a NetworkInfo object containing details about the active networks:
+
+```ts
+interface NetworkInfo {
+  name: string // The human-readable name of the network.
+  networkId: string // The unique identifier for the network (e.g., "mainnet01").
+  url?: string // Optional: The network endpoint URL.
+}
```
alber70g, line 40: See comments on KIP-0039

kip-0039.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+### Success Response
+
+```json
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "result": {
+    "name": "kadena:mainnet01",
```
alber70g, line 105: I wouldn't use `kadena:` in this. The RPC will never return `ethereum:xyz` so it's kinda redundant.
I would consider `networkName` as the display name to show the user in the UI

```suggestion
    "name": "Mainnet",
```

alber70g (changes requested):

kip-0042.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Method Definition
+
+**Method Name:** `kadena_disconnect_v1`
+
+### Request Parameters
+
+```typescript
+interface DisconnectRequest {
+  // TODO: not sure if networkId makes sense here, ecko uses it.
```
alber70g, line 36: I wouldn't include networkId. I think you don't want to disconnect for a single network.

kip-0042.md
```diff
@@ -0,0 +1,135 @@
...
+}
+```
+
+## Response
+
+The method returns a void type, indicating that the disconnection was successful:
+
+```typescript
+type DisconnectResponse = void
+```
```
alber70g, line 47: According to JSON Spec, a `result` is REQUIRED on success. 

```
result
    This member is REQUIRED on success
```

In the example you return `result: {}` which isn't void. Maybe we change the type here to

```suggestion
The method returns a void type, indicating that the disconnection was successful:

```typescript
interface Result {}
```
```

File: kip-0043.md
Comment (alber70g) [inline kip-0043.md:1]: not complete?

**Review by alber70g — COMMENTED** [review]
For each KIP the "Backwards Compatibility" and "Security Considerations" are all the way at the bottom. Maybe it's better to move them above the examples so they're more "part of" the kip

**Comment by javadkh2** [comment]
Can't we just create on KIP for all functions?

File: kip-0037.md [hunk 11]
```diff
@@ -0,0 +1,202 @@
...
+
+# kadena_getAccount_v1 API
+
+## Abstract
+
+We propose the `kadena_getAccount_v1` method to provide an active account known by the wallet. This allows dApps to present the user with the selected account and allow them to use it for creating and signing transactions.
+
+## Motivation
+
+The `kadena_getAccount_v1` method provides a standardized way to retrieve a list of accounts from a Kadena wallet.
```
javadkh2, line 19: does it return a list or only one?

kip-0037.md
```diff
@@ -0,0 +1,202 @@
...
+KIP: '0037'
+Title: kadena_getAccount_v1 API
+Author: Danillo Felixdaal and Bart Huijgen
+Status: Draft
+Type: Standard
+Category: Interface
+Created: 2025-02-17
+---
+
+# kadena_getAccount_v1 API
```
javadkh2, line 11: it seems this method has two use cases so how about creating two methods?
one for getting the active account, and another for retrieving an account by passing public key and contract. 
* kadena_getActiveAccount (no arg)
* kadena_retriveAccount (or kadena_getAccount) ({publicKey, contract})

kip-0037.md
```diff
@@ -0,0 +1,202 @@
...
+interface AccountInfo {
+  accountName: string
+  fungibleContract: string
+  chainAccounts?: Array<{
+    chainId: ChainId
+    guard: {
+      keys: string[]
+      pred: string
+    }
+  }>
```
javadkh2, line 50: This type shows that an account can have a different guard on different chains, which I believe is incorrect
I believe the guard should be directly bound to the account, not the chains.
An account is a combination of an address and a guard. This approach aligns with principal accounts, 
For vanity accounts, if the guard is different, then the wallet and users should treat them as separate accounts.
```suggestion
interface AccountInfo {
  address: string
  fungibleContract: string
  guard: {
      keys: string[]
      pred: string
   }
  availableChains?: Array<{
    chainId: ChainId
  }>
```

kip-0038.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Response
+
+```typescript
+interface AccountInfo {
+  accountName: string
+  fungibleContract: string
+  chainAccounts?: Array<{
+    chainId: ChainId
+    guard: {
```
javadkh2, line 46: same thing about guard here

kip-0039.md
javadkh2, line 47: url can be a list of nodes as well.

kip-0041.md
```diff
@@ -0,0 +1,177 @@
...
+  jsonrpc: '2.0'
+  error: {
+    code: number
+    message: string
+    data?: object
+  }
+}
+```
+
+### Error types
```
javadkh2, line 75: we need an error code if the user rejects the connection request.

nillo:
> Can't we just create on KIP for all functions?

@javadkh2 (@jwiegley john / @alber70g albert decided it was better to split them up)

kip-0037.md
nillo: only one , the getAccount(s) returns a list

kip-0037.md
nillo: @alber70g this feedback was made on your suggestions of passing the publickey, contract as optionals in order to use them as a filter. what you think of @javadkh2 suggestion?

kip-0037.md
nillo: i was under the assumption that a account "CAN" have a different guard on different chains @alber70g @jermaine150

kip-0039.md
nillo: @alber70g javad brings up the same thing we thought about, but we decided to have a single url, want to re-discuss?

kip-0038.md
nillo: @barthuijgen

kip-0037.md
```diff
@@ -0,0 +1,207 @@
...
```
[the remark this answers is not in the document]
nillo: @barthuijgen

kip-0037.md
nillo: @barthuijgen

kip-0039.md
nillo: @barthuijgen

kip-0037.md
alber70g: I believe the most intuitive way to think about accounts is to disconnect them from the guard. A user will think the account is the same until they get the question to define which guard to use. 

Separating accounts based on guard is confusing

kip-0039.md
Randynamic: I would say consistency is key so I agree.

kip-0037.md
nillo: "changed"

kip-0037.md
```diff
@@ -12,123 +12,114 @@ Created: 2025-02-17
...
 **Method Name:** `kadena_getAccount_v1`
 
-### Request Parameters
+#### Parameters
 
-```typescript
-interface GetAccountRequest {
-  account?: {
-    publicKey: string
-    contract: string
```
alber70g: Comment from @javadkh2: when the dapp knows which contract to use, they should be able to select it. It's about the right UX for the user so they don't have to think about it. It can be optional though

kip-0038.md
```diff
@@ -26,81 +37,95 @@ The `kadena_getAccounts_v2` method improved on the previous by allowing extra fe
 
 ### Request Parameters
 
-```typescript
-interface GetAccountsRequest {
-  accounts?: Array<{
-    publicKey: string;
-    contract: string;
```
alber70g: The same thing like above. Also optional. We could add this one later if we want to

alber70g (approved):

kip-0038.md
```diff
@@ -0,0 +1,270 @@
...
+
+No parameters are required.
+
+### Response
+
+The method **MUST** return an array of account objects, even if the wallet only
+supports a single account.
+
+Note that the response can have multiple accounts with the same `accountName`.
+The `guard` is what makes each unique. This allows for dapps to use non-minted
```
LindaOrtega, line 48: > The `guard` is what makes each unique.

Is this statement accurate? 

There could be an account `test1` governed by `guardA` and also an account `test2` that is also governed by `guardA`. 

This could happen when working with vanity accounts or accounts associated with contracts that don't enforce principaled (e.g. that account name must correspond to its associated guard) account names.

kip-0037.md
```diff
@@ -0,0 +1,147 @@
...
+active.
+
+### Response
+
+```typescript
+interface AccountInfo {
+  accountName: string; // The unique identifier for the account.
+  networkId: string; // The unique identifier for the network for this account.
+  contract: string; // Identifier for the fungible token contract.
+  guard: {
```
LindaOrtega, line 44: To get ahead of potential confusion over the definition of "guard" (https://docs.kadena.io/smart-contracts/guards) here:
It might be good to add a note somewhere in the document that we're using the term "guard" here but it focuses mainly on keyset-related guards (keyset ref, single key, multi-key, and w accounts) since that the current ownership model that wallets support.

kip-0037.md
```diff
@@ -0,0 +1,147 @@
...
+No parameters required. The wallet can return any account that is currently
+active.
+
+### Response
+
+```typescript
+interface AccountInfo {
+  accountName: string; // The unique identifier for the account.
+  networkId: string; // The unique identifier for the network for this account.
+  contract: string; // Identifier for the fungible token contract.
```
LindaOrtega, line 43: Should `contract` be optional for accounts that aren't minted yet? Or do wallets always associate accounts with a contract right away?

Also, is it possible for the same `accountName` to be present in multiple contracts? If so, should this field be a list similar to `chainAccounts`?

LindaOrtega (changes requested):
Took a look at KIP 37 and KIP 38 (will get to the rest later). It looks good so far, but I did have some questions and rewording recommendations (see the comments for details).

The main question I'd like to understand better before approving is the following:
- Is an account name always associated with a contract on the wallet side? Particularly, non-minted accounts?

kip-0037.md
```diff
@@ -0,0 +1,147 @@
...
+- **Legacy Methods:** Wallets **MAY** support older methods for retrieving
+  account information alongside `kadena_getAccount_v1`. Legacy methods **MUST**
+  be documented as deprecated and phased out over time.
+
+- **Configuration Consistency:** Implementations **MUST** ensure that the
+  returned account information matches the wallet's active configuration.
+
+## Security Considerations
+
+- **Data Accuracy:** Ensure that the returned account information is accurate
```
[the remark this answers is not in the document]
nillo: The KIP is written for wallets to implement the RPC methods, so the remarks is for the creator of the wallets, i could add, "The wallet creator implementing the RPC methods.." Ensure that the returned account information is accurate before it but that seems kind of redundant to me.

kip-0037.md
```diff
@@ -0,0 +1,147 @@
...
+```typescript
+interface AccountInfo {
+  accountName: string; // The unique identifier for the account.
+  networkId: string; // The unique identifier for the network for this account.
+  contract: string; // Identifier for the fungible token contract.
+  guard: {
+    keys: string[]; // Array of public keys (secret keys are omitted for security).
+    pred: string; // Predicate defining key validation (e.g., "keys-all", "keys-any").
+  };
+  chainAccounts: string[]; // Array of chain IDs where this account exists
```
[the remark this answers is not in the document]
nillo: @barthuijgen @alber70g, i can live with this update thoughts?

kip-0037.md
```diff
@@ -0,0 +1,147 @@
...
+}
```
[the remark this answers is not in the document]
nillo: I don't think this is a assumption that should be made, its either there or its not, if we want it to be there we should enforce it having that said. I seems like a UI thingy to me, the wallet provides the account <- the Dapp itself could provide the ability to add a label to it. Currently none of the wallets support this, so i don't think there is a necessity for this, and should leave labeling to the dapp itself. @barthuijgen @alber70g  thoughts?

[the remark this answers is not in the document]
alber70g: Chainweaver v2 and v3 both have support for aliasses though. If a user is actively using aliasses, it would be nice to have it. I would suggest not to have it nullable, but empty string if it's not there. However current APIs don't support it, correct?

kip-0038.md
```diff
@@ -0,0 +1,270 @@
...
+accounts. When an account isn't minted, the `chainAccounts` field will be an
+empty array.
+
+```typescript
+interface AccountInfo {
+  accountName: string;
+  networkId: string;
+  contract: string;
```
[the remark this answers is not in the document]
alber70g: Yes, when accountName is used for multiple contracts, it's passed multiple times

kip-0038.md
alber70g: In that case they don't have the same `accountName`, right?

kip-0037.md
alber70g: It's any guard honestly. We should maybe extend docs to include all types of guard?

kip-0037.md
alber70g: > Also, is it possible for the same accountName to be present in multiple contracts? If so, should this field be a list similar to chainAccounts?

The wallet only returns the ACTIVE account. Since an account is always tied to a contract, this is not a list

kip-0037.md
barthuijgen: I think guard here makes sense, the other place you mentioned should be changed to refer to multiple keys indeed, which is still one guard.

kip-0037.md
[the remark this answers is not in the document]
barthuijgen: Between the two I think I like `availableChains` the best. `chainAccounts` used to represent something else before, now it indeed does not make much sense.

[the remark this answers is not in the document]
barthuijgen: Me and @nillo aren't sure what to call it, @alber70g what do you prefer?

daplcor:
One thing I'd consider as this will become confusing.  Instead of kadena_getAccounts_v2 API and kadena_getAccounts_v1 API lets not have multiple versions as that would imply updated API features no?
kadena_getAccount_v1 API for single multi lookup
kadena_getAccounts_v1 API for multi wallet lookup.

kip-0038.md
```diff
@@ -0,0 +1,270 @@
...
+1. **Account Information Retrieval:**
+
+   - **MUST** return accurate and up-to-date information for all accounts
+     managed by the wallet.
+   - If a wallet supports only a single active account, the method **MUST**
+     return an array containing that single account.
+   - The returned account information **MUST** include all relevant details as
+     defined in the response interface.
+
+2. **Error Handling:**
```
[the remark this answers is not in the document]
LindaOrtega: no longer relevant because this error type was removed from KIP 38. Resolving comment

kip-0038.md
```diff
@@ -0,0 +1,270 @@
...
+   - **MUST** return a clear and descriptive error message if account
+     information cannot be retrieved.
+   - Errors may result from configuration issues, unsupported networks, or other
```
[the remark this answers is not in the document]
LindaOrtega: No longer relevant after re-write of section:
>    - By default the Internal server error should be used if for any reason account
     information can not be obtained.
>  - Custom errors **SHOULD** be used to give more detail about the cause of the error.

Resolving comment.

kip-0037.md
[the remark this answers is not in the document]
javadkh2: What is the benefit of having chainIds in general?
This will force wallets to query all chains, which makes the response time longer (and also more error-prone) even if the wallet stores data still need to sync data since account creation might happen outside of the wallet with transferCreate.

Also, I can imagine that wallets might just send an empty array to avoid calling all chains, which would then make the response unpredictable and untrustworthy to the end user.

the user can use `accountDiscovery` from client-utils to fetch not only chainId but also the account balance. or if we want the provider returns that we can add another message like kadena_discoverAccount_v1.

kip-0037.md
javadkh2: @alber70g it can't be smart contract related guards since you cant add them to your wallet directly. the valid guards are `Keyset` and `Keyset ref`. the current api only support `keyset` - `{keys: string[];  pred: string}`  . its good to mention it in the KIP and make the guard type more future-proof, allowing support for Keyset Ref later without breaking the API.

LindaOrtega: Is the consensus to add a note in the KIP that the current API only supports `Keyset` guards at the moment?

kip-0038.md
LindaOrtega: The rewording clears it up. Resolving comment.

kip-0037.md
```diff
@@ -0,0 +1,151 @@
...
+No parameters required. The wallet can return any account that is currently
+active.
+
+### Response
+
+```typescript
+interface AccountInfo {
+  accountName: string; // The unique identifier for the account.
+  networkId: string; // The unique identifier for the network for this account.
+  contract: string; // Identifier for the fungible token contract.
```
LindaOrtega, line 43:
```suggestion
  accountName: string; // The unique identifier for the account.
  networkId: string; // The unique identifier for the network for this account.
  contract: string; // Identifier for the fungible token contract. This is a non-empty string since wallets MUST always associate new accounts with a contract.
```

kip-0037.md
LindaOrtega: Added document suggestion related to this here: https://github.com/kadena-io/KIPs/pull/71/files#r2044838690

Resolving this conversation thread

kip-0038.md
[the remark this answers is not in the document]
LindaOrtega: I see this addition to the documentation in previous section that addresses this recommendation:
>Also accounts on different `contract`s can have the same `accountName`, regardless of the guard.

Resolving this conversation thread.

kip-0038.md
```diff
@@ -0,0 +1,270 @@
...
+### Request Parameters
+
+No parameters are required.
+
+### Response
+
+The method **MUST** return an array of account objects, even if the wallet only
+supports a single account.
+
+Note that the response can have multiple accounts with the same `accountName`.
```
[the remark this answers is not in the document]
LindaOrtega: There have been additions to the documentation below this statement that address different scenarios where the same account name can be present (different guards and/or different contracts). Resolving this conversation thread.

kip-0037.md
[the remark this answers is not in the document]
LindaOrtega: @javadkh2 are you proposing that the API for the `kadena_getAccount_v1` be changed so that it no longer returns the chain IDs where the account is present? Have y'all discussed this somewhere else and come to a conclusion?

kip-0037.md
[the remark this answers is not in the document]
LindaOrtega: Did y'all come to a conclusion on whether to include an optional `label` field in this version of the API?

kip-0037.md
```diff
@@ -0,0 +1,151 @@
...
+Type: Standard
+Category: Interface
+Created: 2025-02-17
+---
+
+# kadena_getAccount_v1 API
+
+## Abstract
+
+This KIP standardizes the `kadena_getAccount_v1` method to provide the active
```
LindaOrtega, line 15:
```suggestion
This KIP standardizes the `kadena_getAccount_v1` method, which wallet creators will implement to provide the active
```

kip-0038.md
```diff
@@ -0,0 +1,272 @@
...
+Category: Interface
+Created: 2025-02-17
+---
+
+# kadena_getAccounts_v2 API
+
+## Abstract
+
+This KIP standardizes the `kadena_getAccounts_v2` method, which enables wallets
+to provide a standardized list of accounts they manage. By implementing this
```
LindaOrtega, line 16:
```suggestion
to provide a standardized list of accounts they manage. By using this RPC
```

kip-0038.md
```diff
@@ -0,0 +1,272 @@
...
+method, decentralized applications (dApps) can retrieve and present to users the
+available accounts, allowing users to select the desired account(s) for creating
+and signing transactions.
```
LindaOrtega, line 19:
```suggestion
to provide a standardized list of accounts they manage. By using this RPC
method, decentralized applications (dApps) can retrieve and present to users their
available accounts and, thus, allow users to select the desired account(s) for creating
and signing transactions.
```

kip-0037.md
```diff
@@ -0,0 +1,151 @@
...
+account known by the wallet. This allows dApps to present the user with the
+currently selected account and use it for creating and signing transactions.
```
LindaOrtega, line 17:
```suggestion
This KIP standardizes the `kadena_getAccount_v1` RPC method, which wallet creators will implement to provide the active
account known by the wallet. A dApps can use this method to present the user with the
currently selected account and use it for creating and signing transactions.
```

kip-0037.md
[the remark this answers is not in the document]
LindaOrtega: Yes, adding it here can be redundant. But I added some rewording suggestions to the abstracts of the KIP 37 and 38 that make it a little more clear who will be implementing vs. consuming this RPC method.
- https://github.com/kadena-io/KIPs/pull/71/files#r2044896167
- https://github.com/kadena-io/KIPs/pull/71/files#r2044907844

Resolving this comment thread.

kip-0037.md
LindaOrtega: Adding this rewording for y'all's consideration: https://github.com/kadena-io/KIPs/pull/71/files#r2044923640

kip-0039.md
```diff
@@ -0,0 +1,161 @@
...
+
+Example:
+
+- Wallet returns: https://api.testnet.chainweb.com
+- Dapp can send requests to:
+  https://api.testnet.chainweb.com/chainweb/0.0/testnet04/chain/19/pact/api/v1/local
+
+> [!NOTE]  
+> For more information about Chainweb API see
+> [Chainweb API Documentation](api.chainweb.com)
```
LindaOrtega, line 45: Not redirecting to actual website
```suggestion
> [Chainweb API Documentation](https://api.chainweb.com/openapi/)
```

kip-0039.md
```diff
@@ -0,0 +1,161 @@
...
+by ensuring that network details are provided in a consistent format across
+different wallet implementations. This standardization will:
+
+- Provide reliable network details (such as network name and networkId) to
+  dApps.
+- Simplify integration and reduce potential mismatches between wallet and dApp
+  network settings.
+- Enhance the overall user experience by ensuring clear network context.
+
+The dapp can use the URL to make /local calls from the network.
```
LindaOrtega, line 32: This is the first mention of `URL` in the document. This needs an transition sentence between the previous points (benefits of this method) to the current point (dApps can use the URL).

Maybe something like:
```suggestion
The wallet will also return a list of endpoint URLs for the active network. The dApp can use these URLs to make `/local` calls to the network.
```

and/or move this entire URL section after the Response type definition in the `Specification` section.

(similar comment as in KIP 40)

kip-0039.md
```diff
@@ -0,0 +1,161 @@
...
+
+The wallet returns only the host. The dapp should append networkId and chainId
+to the URL.
```
LindaOrtega, line 35:
```suggestion
The wallet returns only the host. The dApp needs to append the `networkId` and `chainId` to the URL to be able to communicate with the network.
```

kip-0040.md
```diff
@@ -0,0 +1,140 @@
...
+
+Example:
+
+- Wallet returns: https://api.testnet.chainweb.com
+- Dapp can send requests to:
+  https://api.testnet.chainweb.com/chainweb/0.0/testnet04/chain/19/pact/api/v1/local
+
+> [!NOTE]  
+> For more information about Chainweb API see
+> [Chainweb API Documentation](api.chainweb.com)
```
LindaOrtega, line 30: Same comment as KIP 39. This seems out of place in this section. Best to add this after the Response type definition in the `Specification` section.

kip-0040.md
```diff
@@ -0,0 +1,140 @@
...
+
+```ts
+interface NetworkInfo {
+  networkName: string; // The display name of the network (e.g., "mainnet").
+  networkId: string; // The unique identifier for the network (e.g., "mainnet01").
+  url: string[]; // Optional: The root endpoint URL(s) of the network. This should be the base URL for sending commands (e.g., "https://api.chainweb.com") rather than the full command path. Placeholders for chainId should be managed by the client.
+}
+```
+
+The wallet might not want to expose the endpoint to the DApp. In that case the wallet can return an empty array.
```
LindaOrtega, line 50:
```suggestion
If the wallet implementation chooses not to expose the network endpoint, it MUST return an empty array for the `url` field.
```
Same for KIP 39 and KIP 41.

kip-0039.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Specification
+
+### Method Definition
+
+**Method Name:** `kadena_getNetwork_v1`
+
+### Response
+
+The method returns a NetworkInfo object containing details about the active
```
LindaOrtega, line 55:
```suggestion
The method returns a `NetworkInfo` object containing details about the active
```

LindaOrtega (approved):
KIPs 37-41 look technically sound. Approving this with some nitpicks/questions around naming and typing.

The rest of my comments are focused on making sure the KIP is as clear as possible.

kip-0041.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Response
+
+The method returns a `NetworkInfo` object containing details about the newly active network:
+
+```typescript
+interface NetworkInfo {
+  networkName: string; // The display name of the network (e.g., "mainnet").
+  networkId: string; // The unique identifier for the network (e.g., "mainnet01").
+  url: string[]; // Optional: The network endpoint URL, or a list of URLs if multiple nodes are available.
```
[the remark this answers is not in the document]
barthuijgen: should be renamed `urls: []` if its a list

kip-0037.md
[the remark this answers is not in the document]
alber70g: @LindaOrtega I think I would stick to `chainAccounts` as that's also on GraphQL, and since dapps might use that too, it'll be more consistent (https://graph.kadena.network/graphql?query=query+accounts+%7B%0A++fungibleAccount%28accountName%3A%22k%3A554754f48b16df24b552f6832dda090642ed9658559fef9f3ee1bb4637ea7c94%22%29%7B%0A++++accountName%0A++++totalBalance%0A++++chainAccounts+%7B%0A++++++accountName%0A++++++balance%0A++++%7D%0A++%7D%0A%7D)

I believe Tatum also takes that API in their endpoints

emilypi (dismissed):

kip-0037.md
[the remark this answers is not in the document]
javadkh2: @LindaOrtega We didn’t discuss this outside of this thread. I just thought this could add complexity to the wallet side and is also still not reliable on the dApp side. But I can see the benefit of having it if wallets provide it correctly.
For now, we can keep it, but let’s keep an eye on how wallets provide it.

javadkh2 (dismissed):

kip-0037.md
barthuijgen: we're renaming the `guard` field to `keyset`

kip-0037.md
```diff
@@ -0,0 +1,151 @@
...
+
+| Code   | Message                          |
+| ------ | -------------------------------- |
+| -32603 | Internal server error            |
+| -32008 | No account information available |
+
+### Behavior Requirements
+
+1. **Account Retrieval:**
+   - **MUST** return accurate and up-to-date information about the active
```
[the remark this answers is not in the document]
barthuijgen: This spec defines only the communication protocol, these kind of things are up to the wallets to handle. If this requirement suggests otherwise I'd rather remove the requirement from the KIP.

kip-0037.md
```diff
@@ -0,0 +1,151 @@
...
+
+```typescript
+interface AccountInfo {
+  accountName: string; // The unique identifier for the account.
+  networkId: string; // The unique identifier for the network for this account.
+  contract: string; // Identifier for the fungible token contract.
+  guard: {
+    keys: string[]; // Array of public keys (secret keys are omitted for security).
+    pred: string; // Predicate defining key validation (e.g., "keys-all", "keys-any").
+  };
```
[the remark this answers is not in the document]
LindaOrtega: no longer needed with field being renamed to `keyset`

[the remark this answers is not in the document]
LindaOrtega: they renamed the field `keyset`

kip-0041.md
```diff
@@ -0,0 +1,158 @@
...
+```
+
+### Success Response (Single Node)
+
+```json
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "result": {
+    "name": "Mainnet",
```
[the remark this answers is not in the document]
LindaOrtega: Not relevant since KIP 41 was moved from this PR

kip-0041.md
```diff
@@ -0,0 +1,158 @@
...
+Type: Standard
+Category: Interface
+Created: 2025-02-17
+---
+
+# kadena_changeNetwork_v1 API
+
+## Abstract
+
+This KIP standardizes the `kadena_changeNetwork_v1` method for switching the active network within a Kadena wallet. The method updates the wallet’s configuration to the specified network and returns updated network information. This enables dApps to dynamically change networks without a full disconnect, ensuring a seamless user experience.
```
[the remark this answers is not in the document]
LindaOrtega: Not relevant since KIP 41 was moved from this PR

kip-0041.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Behavior Requirements
+
+1. **Network Update:**
+   - **MUST** validate the provided `networkId` against the wallet’s supported networks.
+   - **MUST** update the wallet configuration to use the specified network.
+   - **MUST** return the updated network details in the response.
+2. **Data Accuracy:**
+   - **MUST** ensure that the returned `NetworkInfo` accurately reflects the wallet’s active configuration.
+3. **Security Considerations:**
```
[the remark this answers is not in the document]
LindaOrtega: Not relevant since KIP 41 was moved from this PR

kip-0041.md
```diff
@@ -0,0 +1,158 @@
...
+
+```typescript
+interface NetworkInfo {
+  networkName: string; // The display name of the network (e.g., "mainnet").
+  networkId: string; // The unique identifier for the network (e.g., "mainnet01").
+  url: string[]; // Optional: The network endpoint URL, or a list of URLs if multiple nodes are available.
+}
+```
+
+The wallet might not want to expose the endpoint to the DApp. In that case the wallet can return an empty array.
```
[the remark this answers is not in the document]
LindaOrtega: Not relevant since KIP 41 was moved from this PR

kip-0040.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Response
+
+The method returns an Array of `NetworkInfo` objects, containing details about the active networks:
+
+```ts
+interface NetworkInfo {
+  networkName: string; // The display name of the network (e.g., "mainnet").
+  networkId: string; // The unique identifier for the network (e.g., "mainnet01").
+  url: string[]; // Optional: The root endpoint URL(s) of the network. This should be the base URL for sending commands (e.g., "https://api.chainweb.com") rather than the full command path. Placeholders for chainId should be managed by the client.
```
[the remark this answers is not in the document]
LindaOrtega: This was moved out of the comments and into the markdown

barthuijgen (approved):

edmundnoble (approved):

LindaOrtega (approved):

alber70g (approved):

